PR #21, opened 2021-11-02
feat: H>ZZ* pipeline notebook polishing and small fixes

### LINKED ISSUE #20: Save visualizations as PDF and PNG post axis and label manipulation
Let me first say that @alexander-held and @oshadura have done a super impressive job with all the notebook demos and this Issue should be considered a super small nitpick that is an "if there is time" thing to consider. For the [current state of the HZZ_analysis_pipeline notebook](https://github.com/iris-hep/analysis-grand-challenge/blob/f71b76f316538545c72bea41602c58d0fcf2d5f6/workshops/agctools2021/HZZ_analysis_pipeline/HZZ_analysis_pipeline.ipynb) everything inside the notebook looks super good. :tada: 

What would be nice though is if all visualizations could be:
- saved under the `figures` directory
- as both PDF and PNG file formats ([with `facecolor="white"`](https://twitter.com/HEPfeickert/status/1319436023533834240?s=20))
- have any axis label manipulations performed after initial figure generation saved out as well.

* At the moment, the `mplhep` and `hist` plot in cell 9 is not saved out as a figure under the `figures` directory.
* The `cabinetry` pulls and correlation matrix (cell 15 and 16) plots look great, but are saved (by default) as PDF only, so it would be good if they could be saved as PNG as well.
* The final 4 lepton invariant mass plot (cell 17) looks great as well in the rendered notebook

download

however, the PDF version that is saved out to the `figures` directory

without_edits

doesn't have the label modifications of

```python
# modify x-axis label
figure_dict[0]["figure"].axes[1].set_xlabel("m4l [GeV]")
```

applied to it. It would be great if the saved versions and the displayed versions in the notebook could be identical.

Like I said, all super small nitpicks that aren't make or break for the demos. I can't promise that I'll have time to address these tonight, but if I do have time I'll open a PR for these, but if someone else is feeling bored please feel free to beat me to it. :+1:

alexander-held:
This adds further comments to the H>ZZ* pipeline notebook, including a new section and figure at the end. Various small formatting fixes are included, and the git submodules are changed to use `https` to work on coffea-casa.

A new utility function is used to save figures in both `.pdf` and `.png` formats with correct facecolor. Thanks @matthewfeickert for suggesting this!

resolves #20


matthewfeickert:
Awesome! Thanks for being so fast on acting on this @alexander-held, as I know it is already quite late for you.

PR #67, opened 2022-07-07
add requirements.txt for cms-open-data-ttbar notebook

davekch:
added a `requirements.txt` so that people can quickly setup an environment to run the cms open data ttbar notebook


analyses/cms-open-data-ttbar/requirements.txt
```diff
@@ -0,0 +1,146 @@
...
+numpy==1.21.6
+packaging==21.3
+pandas==1.4.3
+pandocfilters==1.5.0
+parso==0.8.3
+partd==1.2.0
+pexpect==4.8.0
+pickleshare==0.7.5
+Pillow==9.2.0
+pkg_resources==0.0.0
```
alexander-held, line 94: I think this shouldn't be in here, at least it causes issues when trying to `python -m pip install -r requirements.txt`:
```
ERROR: No matching distribution found for pkg_resources==0.0.0
```

alexander-held (commented):
Thanks a lot @davekch for this PR! It is very useful to have this available in order to quickly set up the right environment.

analyses/cms-open-data-ttbar/requirements.txt
davekch: that's interesting, I don't get this error. Using Python 3.8.10

alexander-held: The behavior might be OS-specific: https://stackoverflow.com/questions/39577984/what-is-pkg-resources-0-0-0-in-output-of-pip-freeze-command. As far as I can tell from this it should be safe to remove it.

davekch: interesting. I removed it

alexander-held (approved):
Thanks again! Looks all good.

PR #72, opened 2022-08-04
docs: add license

alexander-held:
We were missing a license, fixing this by adding the MIT license to the repository.


oshadura (approved):
LGTM!

PR #73, opened 2022-08-04
feat: add small tests

alexander-held:
Adding a new folder to track small tests for integration testing purposes of (parts of) the AGC pipeline:

- minimal coffea + ServiceX pipeline
- various ServiceX queries
- `servicex-databinder` example


oshadura:
@alexander-held let's merge it, I don't see any issue to have tests available in the repo.

PR #84, opened 2022-10-17
feat: databinder to coffea interface

alexander-held:
Use `servicex-databinder` to only get a list of parquet file URLs, then feed those to coffea. Also improving func_adl query to use proper event selection (needs to be relaxed for systematics acceptance).


alexander-held:
This can presumably be superseded by an implementation similar to #96.

alexander-held:
Superseded by #103.

PR #86, opened 2022-10-20
feat: I/O testing via awkward.materialized

alexander-held:
Removing all processing from the coffea processor, only do I/O via awkard.materialized.


alexander-held:
Added this directly to #85 instead, so this can be closed.

PR #89, opened 2022-11-16
[pip] Update AGC related packages to latest versions

oshadura:


alexander-held (approved):
I tested this locally and it works with `IterativeExecutor` and `DaskExecutor`. `FuturesExecutor` fails locally but looks unrelated. Thanks!

PR #92, opened 2022-11-21
feat: use jupytext for demonstrator notebooks

ekauffma:
Paired markdown and py:percent files are added for the ATLAS analysis notebook (via Jupytext). A pre-commit config file is also included. In order to run using the pre-commit hook (which ensures all files are synced via Jupytext when committing), the developer must make sure the pre-commit hook is installed by running "pre-commit install" while in the repository. Instructions for this are also included in Markdown in the file itself.


ekauffma:
> Nice. :) You have the pre-commit hook, but if you also wanted to have files be autopaired you could also add [the following](https://github.com/matthewfeickert/pythia-python/blob/0f5988b60978647fca23ecfe2af22f0de8a7e622/pyproject.toml) to a `pyproject.toml` in the top level directory
> 
> ```toml
> [tool.jupytext]
> # Always pair ipynb notebooks to py:percent files
> formats = ["ipynb", "py:percent"]
> notebook_metadata_filter = "all,-jupytext.text_representation.jupytext_version"
> ```
> 
> That may or may not be of interest though.

Thanks! Yes I did want to do this as well, although I think that @oshadura wanted to avoid creating paired files for the notebooks in the `workshops/` directory.. I think I remember some way to do this though.. will look into it!

matthewfeickert:
Yeah, you can't do exclude patterns (AFAIK) so you either would need to specify the directories that you _do_ want paired notebooks to be generated for in the config (c.f. https://github.com/mwouts/jupytext/issues/1033) or have a `jupytext.toml` per directory.

analyses/cms-open-data-ttbar/coffea.ipynb
```diff
@@ -7,6 +7,8 @@
    "source": [
     "# CMS Open Data $t\\bar{t}$: from data delivery to statistical inference\n",
     "\n",
+    "<span style=\"color:red\"> *NOTE: this is an old version of the CMS ttbar analysis notebook. Development on this notebook has moved to `analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.ipynb`.* </span>\n",
```
alexander-held, line 10: Could we please remove anything below this in the notebook to not keep duplicated code around?

analyses/cms-open-data-ttbar/README.md
```diff
@@ -0,0 +1,23 @@
+#### Instructions for paired notebook
+
+If you only care about running the `ttbar_analysis_pipeline.ipynb` notebook, you can completely ignore the `ttbar_analysis_pipeline.py` file.
+
+This notebook (`ttbar_analysis_pipeline.ipynb`) is paired to the file `ttbar_analysis_pipeline.py` via Jupytext (https://jupytext.readthedocs.io/en/latest/). Using `git diff` with this file instead of the `.ipynb` file is much simpler, as you don't have to deal with notebook metadata or output images. However, in order for the notebook output to be preserved, the notebook still needs to be version controlled. It is ideal to run `git diff` with the option `-- . ':(exclude)*.ipynb'`, so that `.ipynb` files are ignored. 
+
+The `.py` file can also be run as a Python script.
```
alexander-held, line 7: Here we (currently) will still have the problem with the `async` code, right?

analyses/cms-open-data-ttbar/coffea.ipynb
ekauffma: Done!

analyses/cms-open-data-ttbar/README.md
ekauffma: What problem? I am not sure what you are referring to

alexander-held: This is ServiceX-specific, the `async def produce_all_the_histograms()` section is needed to run successfully as a script (so the notebook version needs to be commented out and that other piece needs to be enabled).

This issue will go away via #96, so I realize now that it won't matter here.

alexander-held (approved):
Looks good to me, thank you! I re-synced the ttbar notebook with the current HEAD version to avoid changing anything else in this PR.

matthewfeickert:
> when does this auto-pairing happen? I am trying to understand the following scenarios:

Sorry @alexander-held missed these. The auto-pairing happens at creation time if you're using a `pyproject.toml` config or a `jupytext.toml` config. So if you created an `example.ipynb` and then saved it, jupytext would automatically create a paired `example.py` at save.

> 1. A pair of .py and .ipynb already exist in the repository and a user updates the notebook. The syncing is automatic via the `pre-commit` hook I believe?

Hm that's a good question about the pre-commit hook. I assume yes, but I'd have to try it myself (I hadn't used the pre-commit hook yet). I'm assuming that @ekauffma knows given they added the pre-commit hook. I can try this out on another project though tonight at report back.

> 2. A user creates and commits a new .ipynb notebook. Does the `pre-commit` hook create a synced .py? Do we need the `pyproject.toml` for that?

I would assume that here the `pre-commit` hook would do this, but as mentioned above if you either manually select a notebook to be paired with a `py:percent` format or specify this in a jupytext config then it will sync on save.

PR #96, opened 2022-12-15
feat: serving ROOT file URIs with ServiceX in HZZ notebook (AGC demo day)

talvandaalen:
Changed the ATLAS OpenData HZZ analysis notebook to use standalone instances of `servicex` and `coffea`, by using new sync functions in the `servicex` frontend to retrieve the URIs of queried files. These are then passed to standalone `coffea`, and the analysis progresses similarly to before.


analyses/atlas-open-data-hzz/HZZ_analysis_pipeline.ipynb
```diff
@@ -278,24 +519,33 @@
...
+    "\n",
+    "        # construct lepton four-vectors\n",
+    "        # TODO: might be able to replace by schema\n",
+    "        leptons = ak.zip(\n",
+    "            {\"pt\": events.lep_pt,\n",
+    "             \"eta\": events.lep_eta,\n",
+    "             \"phi\": events.lep_phi,\n",
+    "             \"energy\": events.lep_energy},\n",
+    "            with_name=\"Momentum4D\",\n",
+    "        )\n",
```
alexander-held, line 548: I'm not sure how this ran without a `vector.register_awkward()` in the past? I'll add it to the processor but without that I cannot get this to run.

alexander-held:
Note: this reads a ServiceX backend with version 1.1.3 or later to run.

alexander-held (approved):

PR #102, opened 2023-02-16
feat: switch to NanoAOD input files for AGC CMS ttbar notebook

### LINKED ISSUE #63: Switch to nanoAOD inputs for CMS Open Data pipeline demonstration
Follow-up to #31: Once available, switch from ntuple inputs to nanoAOD inputs for the CMS Open Data pipeline demonstration.

### Requires
- [ ] availability of nanoAODs for 2015 CMS Open Data, or a miniAOD -> nanoAOD conversion tool
- [ ] ...

ekauffma:
- Branch and tree names were altered accordingly,
- Different branches are selected for the only_do_IO method in the coffea processor
- Table with file sizes is changed to reflect the new files

resolves #63


analyses/cms-open-data-ttbar/ntuples_nanoaod.json
```diff
@@ -0,0 +1,3256 @@
+{
```
alexander-held, line 1: Let's call this file "nanoaod_inputs.json" or something like that, the "ntuples" name might be slightly misleading.

ekauffma: Done!

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -78,53 +79,37 @@
...
-PIPELINE = "coffea"
+PIPELINE = "servicex_processor"
 
 # enable Dask (may not work yet in combination with ServiceX outside of coffea-casa)
-USE_DASK = True
+USE_DASK = False
 
 # ServiceX behavior: ignore cache with repeated queries
-SERVICEX_IGNORE_CACHE = False
+SERVICEX_IGNORE_CACHE = True
```
alexander-held, line 112: Same here, using the cache is a more reasonable default.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -78,53 +79,37 @@
...
 # pipeline to use:
 # - "coffea" for pure coffea setup
 # - "servicex_processor" for coffea with ServiceX processor
 # - "servicex_databinder" for downloading query output and subsequent standalone coffea
-PIPELINE = "coffea"
+PIPELINE = "servicex_processor"
 
 # enable Dask (may not work yet in combination with ServiceX outside of coffea-casa)
-USE_DASK = True
+USE_DASK = False
```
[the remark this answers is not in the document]
ekauffma: Ok let me re-run the notebook real quick then!

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -198,22 +183,45 @@ def __init__(self, disable_processing, io_file_percent):
...
-            if self.io_file_percent not in [4, 15, 25, 50]:
-                raise NotImplementedError("supported values for I/O percentage are 4, 15, 25, 50")
-            if self.io_file_percent >= 15:
-                branches_to_read += ["trigobj_e"]
-            if self.io_file_percent >= 25:
-                branches_to_read += ["trigobj_pt"]
-            if self.io_file_percent >= 50:
-                branches_to_read += ["trigobj_eta", "trigobj_phi", "jet_px", "jet_py", "jet_pz", "jet_ch"]
-
+        # standard AGC branches cover 2.7% of the data
```
alexander-held, line 186: It's somewhat counterintuitive to me that we read a smaller fraction of data using the same branches, even though the file size also decreased. I imagine it must be that the number of jets / leptons per events are smaller in the NanoAODs since the previous ntuples probably had less selection requirements applied to them already?

ekauffma: Maybe. Weren't there also many duplicate counts branches in the old files? Maybe this is related.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -416,35 +384,23 @@ def behavior(self):
...
+                 .Where(lambda e: e.Jet_pt.Where(lambda pt: pt > 25).Count() >= 4)\
+                 .Where(lambda g: {"pt": g.Jet_pt, 
+                                   "btagCSVV2": g.Jet_btagCSVV2}.Zip().Where(lambda jet: 
+                                                                             jet.btagCSVV2 >= 0.5 
+                                                                             and jet.pt > 25).Count() >= 1)\
+                 .Select(lambda f: {"Electron_pt": f.Electron_pt,
+                                    "Electron_mass": f.Electron_mass,
+                                    "Muon_pt": f.Muon_pt,
+                                    "Muon_mass": f.Muon_mass,
+                                    "Jet_mass": f.Jet_mass,
```
alexander-held, line 398: Do we need the masses here specifically or does the energy also work?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -416,35 +384,23 @@ def behavior(self):
...
-        )
+    return source.Where(lambda e: e.Electron_pt.Where(lambda pt: pt > 25).Count() 
+                        + e.Muon_pt.Where(lambda pt: pt > 25).Count() == 1)\
+                 .Where(lambda e: e.Jet_pt.Where(lambda pt: pt > 25).Count() >= 4)\
+                 .Where(lambda g: {"pt": g.Jet_pt, 
+                                   "btagCSVV2": g.Jet_btagCSVV2}.Zip().Where(lambda jet: 
+                                                                             jet.btagCSVV2 >= 0.5 
+                                                                             and jet.pt > 25).Count() >= 1)\
+                 .Select(lambda f: {"Electron_pt": f.Electron_pt,
+                                    "Electron_mass": f.Electron_mass,
```
alexander-held, line 395: I forgot why we had to previously use the energy (I think something broke without it, perhaps it won't anymore), do we actually need the mass here at all?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: Energy did not work. Maybe because the energy is not a branch in the nanoAOD (It is calculated when running the NanoAODSchema)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: No we do not

alexander-held (commented):
Thanks a lot for preparing this, the jupytext use is already paying off for this PR alone. Nice to see that the changes required overall really aren't all that many!

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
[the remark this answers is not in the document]
ekauffma: Done

alexander-held:
> They are pre-converted from miniAOD files into ntuple format, similar to nanoAODs.

We need to change that line into something like

> They were originally published in miniAOD format.
For the purpose of this demonstration these files were pre-converted into nanoAOD format and the nanoAOD files are used as input here.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -234,10 +258,10 @@ def process(self, events):
             # very very loosely based on https://arxiv.org/abs/2006.13076
 
             # pT > 25 GeV for leptons & jets
-            selected_electrons = events.electron[events.electron.pt > 25]
-            selected_muons = events.muon[events.muon.pt > 25]
-            jet_filter = events.jet.pt * events[pt_var] > 25  # pT > 25 GeV for jets (scaled by systematic variations)
```
alexander-held, line 239: Let's keep this old comment here, it will be superseded with better systematics handling in a subsequent version but otherwise I think this `events[pt_var]` approach might look quite confusing.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -392,35 +376,21 @@ def behavior(self):
...
-            }
-        )
+    return source.Where(lambda e: e.Electron_pt.Where(lambda pt: pt > 25).Count() 
+                        + e.Muon_pt.Where(lambda pt: pt > 25).Count() == 1)\
+                 .Where(lambda e: e.Jet_pt.Where(lambda pt: pt > 25).Count() >= 4)\
+                 .Where(lambda g: {"pt": g.Jet_pt, 
+                                   "btagCSVV2": g.Jet_btagCSVV2}.Zip().Where(lambda jet: 
+                                                                             jet.btagCSVV2 >= 0.5 
+                                                                             and jet.pt > 25).Count() >= 1)\
+                 .Select(lambda f: {"Electron_pt": f.Electron_pt,
```
alexander-held, line 386: Regarding the names used here for the event: I remember that there was a bug at some point where using `e` for all of them didn't work, but as far as I know that is resolved (and `e` everywhere works fine). Could we please either use `e` everywhere or do something like `e`, `f`, `g`, `h` or so? At the moment it looks like there might be some purpose to `e`, `e`, `g`, `f`, but if there is I don't understand it.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -76,16 +76,17 @@
...
 # | --- | --- | --- |
-# | `1` | 9 | 16.3 GB |
-# | `5` | 45 | 81.7 GB |
-# | `10` | 86 | 157 GB |
-# | `50` | 357 | 678 GB |
-# | `100` | 590 | 1.09 TB |
-# | `500` | 1542 | 2.58 TB |
-# | `1000` | 2249 | 3.57 TB |
-# | `-1` | 2269 | 3.59 TB |
+# | `1` | 12 | 25.1 GB |
```
alexander-held, line 79: Did the splitting into categories change, or why is the minimum number of files for setting `1` changing from 9 to 12?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: I will change to the latter option, because when everything is `e`, I get `ValueError: Found multiple calls to on same line - split the calls across lines or change lambda argument names so they are different.`

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: for single_top_tW, for nanoAOD I have scaledown, scaleup, and DS in addition to nominal. Should I remove these?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
alexander-held: found the issue: https://github.com/iris-hep/func_adl/issues/115, might be resolved by a newer version (but changing name is also fine)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
alexander-held: Yes, let's remove these as they are currently not being used. I'll track the paths via #118 but this is low priority to include again in the pipeline.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
alexander-held: will investigate more closely in a follow-up issue or PR

alexander-held (approved):
All looks good, thank you!

PR #103, opened 2023-02-20
feat: serve ROOT file URIs with ServiceX for CMS ttbar notebook

### LINKED ISSUE #93: SyntaxError: 'await' outside function
If coffea.ipynb os converted to a script using nbconvert, the resulting script will abort with the message in the title, due to the statement

all_histograms = await utils.produce_all_histograms(...)

### LINKED ISSUE #100: Port ServiceX serving ROOT URIs to ttbar notebook
Port #96 to the CMS ttbar notebook.

cc @ekauffma

ekauffma:
Updated CMS ttbar pipeline to match the changes in the ATLAS analysis here: 81496803ef2f544c102a913ab3f4afeddf6248ce 

This effectively ports over #96. Resolves #100. Resolves #93 (async interface is no longer used after this change here).


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -448,107 +412,92 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
+exec_time = time.monotonic() - t0
+
+all_histograms = all_histograms["hist"]
 
 print(f"\nexecution took {exec_time:.2f} seconds")
 
 # %%
 # track metrics for pure coffea setups
-if PIPELINE == "coffea":
+if not USE_SERVICEX:
```
alexander-held, line 496: We don't need this check anymore, it previously existed since the ServiceX processor did not write out metrics. Now that we're using the same `coffea` approach either way, we always get metrics. Let's add the `USE_SERVICEX` setting to the metrics as well for tracking purposes.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -88,39 +87,17 @@
...
 # input files per process, set to e.g. 10 (smaller number = faster)
 N_FILES_MAX_PER_SAMPLE = 5
 
-# pipeline to use:
-# - "coffea" for pure coffea setup
-# - "servicex_processor" for coffea with ServiceX processor
-# - "servicex_databinder" for downloading query output and subsequent standalone coffea
-PIPELINE = "coffea"
+# enable ServiceX
+USE_SERVICEX = False
```
alexander-held, line 98: This is pretty minor, but perhaps group this together with the cache setting since they're both about ServiceX.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -448,107 +412,92 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
+
+    fileset_queried = {}
+
+    for process in fileset.keys():
+        ds = ServiceXDataset(fileset[process]['files'], backend_name="uproot", ignore_cache=SERVICEX_IGNORE_CACHE)
+        files = ds.get_data_rootfiles_uri(query, as_signed_url=True)
+
+        fileset_queried[process] = {"files": [f.url for f in files],
+                                    "metadata": fileset[process]['metadata']
+                                   }
```
alexander-held, line 458: I missed this in #96, but is there any reason we cannot directly override `fileset_queried[process]["files"]` and thereby avoid having to worry about the metadata? That should leave whatever else is in the dictionary unchanged (in case we adjust that in the future).

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -394,25 +371,12 @@ def behavior(self):
...
-# Here, we gather all the required information about the files we want to process: paths to the files and asociated metadata.
-
-# %% tags=[]
-fileset = utils.construct_fileset(N_FILES_MAX_PER_SAMPLE, use_xcache=False, af_name=AF_NAME)  # local files on /data for ssl-dev
-
-print(f"processes in fileset: {list(fileset.keys())}")
-print(f"\nexample of information in fileset:\n{{\n  'files': [{fileset['ttbar__nominal']['files'][0]}, ...],")
-print(f"  'metadata': {fileset['ttbar__nominal']['metadata']}\n}}")
-
-
```
alexander-held, line 409: Is there a reason the order flipped with the query and the fileset construction?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -448,107 +412,92 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
+
+    for process in fileset.keys():
+        ds = ServiceXDataset(fileset[process]['files'], backend_name="uproot", ignore_cache=SERVICEX_IGNORE_CACHE)
+        files = ds.get_data_rootfiles_uri(query, as_signed_url=True)
+
+        fileset_queried[process] = {"files": [f.url for f in files],
+                                    "metadata": fileset[process]['metadata']
+                                   }
+
     print(f"execution took {time.time() - t0:.2f} seconds")
```
alexander-held, line 460: Let's make this a bit more informative with "ServiceX data delivery" or so.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
alexander-held: I missed that `fileset_queried` is a new dictionary, what I had in mind was overriding `fileset` since that happens a few cells down anyway.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: I flipped it because I put the servicex step right after fileset construction. But I think it makes sense to keep the servicex stuff together after fileset construction so I will change it.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -448,107 +412,92 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
-    raise NotImplementedError("further processing of this method is not currently implemented")
+if USE_DASK:
+    executor = processor.DaskExecutor(client=utils.get_client(AF))
+else:
+    executor = processor.FuturesExecutor(workers=NUM_CORES)
+        
+run = processor.Runner(executor=executor, schema=AGCSchema, savemetrics=True, metadata_cache={}, chunksize=CHUNKSIZE)
+
+if USE_SERVICEX:
+    treename = "servicex"
```
alexander-held, line 478: I asked on Slack whether there's a way we can avoid this part and directly set the name in the query, which would simplify this.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: Addressed in [d87e4bc](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/d87e4bcb704e9e57af7d68b3f1c60cd567705168)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: Fixed in [9587d6c](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/9587d6c0f7766965c0f95282c5a78fbcf1c14001)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: Fixed in [9587d6c](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/9587d6c0f7766965c0f95282c5a78fbcf1c14001)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: Addressed in [9587d6c](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/9587d6c0f7766965c0f95282c5a78fbcf1c14001)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: Addressed in [d2ab835](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/d2ab83532ae72e5ecd07f3f0702a442f26f95842)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -448,121 +425,98 @@ def get_query(source: ObjectStream) -> ObjectStream:
 
 
 # %% [markdown]
-# ### Standalone ServiceX for subsequent `coffea` processing
+# ### Caching the queried datasets with `ServiceX`
 #
-# Using `servicex-databinder`, we can execute a query and download the output.
-# As the files are currently accessible through `rucio` only with ATLAS credentials, you need to use an ATLAS ServiceX instance to run this (for example via the UChicago coffea-casa analysis facility).
+# Using the queries created with `func_adl`, we are using `ServiceX` to read the ATLAS Open Data files to build cached files with only the specific event information as dictated by the query.
```
alexander-held, line 430: > ATLAS Open Data

should be CMS

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -448,121 +425,98 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
+    # dummy dataset on which to generate the query
+    dummy_ds = ServiceXSourceUpROOT("cernopendata://dummy", "events", backend_name="uproot")
+
+    # tell low-level infrastructure not to contact ServiceX yet, only to
+    # return the qastle string it would have sent
+    dummy_ds.return_qastle = True
+
+    # create the query
+    query = get_query(dummy_ds)
+    query = query.value()
```
alexander-held, line 444: Might as well do this in one line
```python
query = get_query(dummy_ds).value()
```
as I'm not aware of anything else we'd do with this intermediate object.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -448,121 +425,98 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
+    print(f"ServiceX data delivery took {time.time() - t0:.2f} seconds")
 
 # %% [markdown]
 # ### Execute the data delivery pipeline
 #
-# What happens here depends on the configuration setting for `PIPELINE`:
-# - when set to `servicex_processor`, ServiceX will feed columns to `coffea` processors, which will asynchronously process them and accumulate the output histograms,
-# - when set to `coffea`, processing will happen with pure `coffea`,
-# - if `PIPELINE` was set to `servicex_databinder`, the input data has already been pre-processed and will be processed further with `coffea`.
+# What happens here depends on the flag `USE_SERVICEX`. If set to true, the processor is run on the data previously gathered by  ServiceX, then will gather output histograms.
```
alexander-held, line 462: double space before ServiceX

ekauffma: [8a8b37e](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/8a8b37e5686fd0104fb841a442f6c37a4b619fd4)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: [b1561e3](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/b1561e3a281d83fb5803928d8e6c5d46502fb609)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: [66ff1ba](https://github.com/iris-hep/analysis-grand-challenge/pull/103/commits/66ff1ba7404a27006989cff15e0b71847c587648)

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
alexander-held: Will resolve here and we can address this in a different PR if there's a better solution.

alexander-held:
I just noticed that the transforms show up as "Untitled" on the dashboard, we should name them according to the process (like they were previously). I think we passed that information through to the utils module somehow.

alexander-held (approved):
Looks good, thank you!

PR #104, opened 2023-02-21
feat: ATLAS HZZ ServiceX transform names

ekauffma:
Adds names to the transforms performed by ServiceX.


alexander-held (approved):
Works as expected, thanks!

PR #107, opened 2023-02-22
feat: combine ServiceX transforms for CMS ttbar notebook

alexander-held:
This follows @ekauffma's idea to group transforms together for more efficient processing (now that we are no longer running them async) and is based on his first implementation.

The custom `ServiceXDatasetGroup` class is mostly meant as an exploration for how well such an API might work. In the future this should be upstreamed into the ServiceX client library.


analyses/cms-open-data-ttbar/utils/__init__.py
```diff
@@ -119,4 +120,29 @@ def save_histograms(all_histograms, fileset, filename):
...
+            filelist += fileset[process]["files"]  # add all files together into list for processing
+
+        self.ds = ServiceXDataset(filelist, backend_name=backend_name, ignore_cache=ignore_cache)
+
+    def get_data_rootfiles_uri(self, query, as_signed_url=True, title="Untitled"):
+        all_files = self.ds.get_data_rootfiles_uri(query, as_signed_url=as_signed_url, title=title)
+        files_per_process = {}
+        for process in self.fileset:
+            # processdict for translation using file name of parent
+            files_per_process.update({process: [f for f in all_files if self.filename_to_process[f.file.split(":")[-2]] == process]})
```
alexander-held, line 147: A more generic implementation would need to translate the input file name into the format used by `.file` (e.g. `/` -> `:` substitutions).

alexander-held:
See also https://github.com/iris-hep/func_adl_servicex/issues/56.

PR #109, opened 2023-03-07
Feat/servicex dataset group

ekauffma:
Changed process dictionary method of matching ServiceX output files to input files to method which matches filename strings exactly after replacing "/" with ":". Avoids dependence on file-naming convention.


alexander-held (approved):
Thanks! This is a lot better than what we had previously.

Issue #91, opened 2022-11-16
Local use of `FuturesExecutor`


alexander-held:
Testing #89 revealed that `FuturesExecutor` locally fails with
```pytb
AttributeError: Can't get attribute 'AGCSchema' on <module '__main__' (built-in)>
```
On coffea-casa at UNL and at SSL this works fine.

This can be avoided by converting the notebook
```bash
jupyter nbconvert --to script coffea.ipynb
mv coffea.py agc.py
```
where renaming is important as the name `coffea` collides with the module name. This then results in
```pytb
AttributeError: module 'vector' has no attribute 'behavior'
```
which can be solved by removing `import vector; vector.register_awkward() ` which might no longer be needed after the removal of `vector` to construct 4-vectors manually for ServiceX.

We should track down why exactly these things fail to provide a more general solution.

alexander-held:
This was effectively solved via #103, which removes the need for ServiceX processors in coffea that used `vector` to re-build a four-vector object. The prior adoption of `jupytext` also removes the need to convert / rename files.

alexander-held:
There are two things that can go wrong:
1) Using a script, but the futures executor is not guarded (https://stackoverflow.com/a/15918742). Avoid with a guard:
```python
from coffea import processor
from coffea.nanoevents import BaseSchema

class MyProcessor(processor.ProcessorABC):
    def __init__(self):
        pass

    def process(self, events):
        return {}

    def postprocess(self, accumulator):
        pass


if __name__ == "__main__":
    fileset = {"ZZ to 4mu": ["root://eospublic.cern.ch//eos/root-eos/cms_opendata_2012_nanoaod/ZZTo4mu.root"]}

    executor = processor.FuturesExecutor(workers=1)
    run = processor.Runner(executor=executor, schema=BaseSchema, maxchunks=1)
    run(fileset, "Events", processor_instance=MyProcessor())
```

2) Using a dynamically defined class as schema. This works fine on coffea-casa, not locally on macOS. With a class like
```python
class AGCSchema(BaseSchema):
    pass
```
used as schema, the pickling fails. This is only a problem in notebooks. It can be avoided by implementing the class externally and importing it instead.

PR #112, opened 2023-04-04
build: update versions of dependencies from CMS ttbar setup

alexander-held:
Updating `requirements.txt`. Tested and generally works fine, some issues with Python futures on macOS but that is probably an independent problem.

For future reference, the base libraries to build this can be obtained via:
```bash
pip install jupyter awkward bokeh cabinetry coffea dask distributed func-adl func-adl-servicex hist jupyter matplotlib numpy servicex uproot
```


alexander-held:
The local issues with futures are unrelated, see https://github.com/iris-hep/analysis-grand-challenge/issues/91#issuecomment-1496268652.

PR #117, opened 2023-04-13
fix: add axis labels and binning to cabinetry config file

ekauffma:
Cabinetry pre-fit and post-fit plots now have appropriate axis labels and bin values after changing cabinetry config file.


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -613,8 +613,10 @@ def get_query(source: ObjectStream) -> ObjectStream:
 # The binning here corresponds to the binning used for the fit.
 
 # %%
+plt.rcParams["font.size"] = 10
```
alexander-held, line 616: Let's move this to `utils`, there is a `set_style()` function for that.

ekauffma: I wanted to make the font size smaller than the one set in utils so there is no overlap with the legend. should I make this an argument in the `set_style()` function?

alexander-held: I am not actually completely sure whether putting it in `set_style()` works like I thought due to https://github.com/scikit-hep/cabinetry/issues/381#issuecomment-1335677155: perhaps the `cabinetry`-internal `mpl.style.use` call would override this value. That call is within a function called from `cabinetry.visualize.data_mc` though, so I think if your current approach works, moving it to `set_style()` would work too.

Let's try whether making the `utils` just generally font size 10 is ok, I imagine for the other affected plots that is not too big of an issue. Otherwise we can leave this out here, I would just like to factor out as much of the styling as possible to keep the notebook more focused on the pipeline.

ekauffma: Yep, changing all to 10 works fine.

alexander-held: Great, let's go with that. This also means that `mpl.style.use` probably doesn't quite work like I thought it would.

PR #119, opened 2023-04-17
feat: adopt correctionlib for applying systematic variations

### LINKED ISSUE #110: Structure for processor for correctionlib adoption / refactor
- `event` object contains event data
- call to `correctionlib` to get `event_nominal` version with all objects calibrated to their nominal state
- loop over all systematic variations that change object kinematics (+ nominal)
    - figure out what objects changed, vary those objects that did change via `correctionlib` -> `event_for_this_systematic`
    - run event selection + observable calculation (*1)
    - fill histograms
        - if nominal: also fill all weight variations (in another loop over weight-based systematics), weights obtained via correctionlib

(*1): this does not automatically avoid re-calculation of quantities that do not change (e.g. those derived from jets for muon systematic variations)

cc @andrew42

see also Lindsey's talk at the last AGC demo day https://indico.cern.ch/event/1232470/#4-awkward-dask-integration-int

Andrew42:
Implements the use of the `correctionlib` library for applying systematic corrections in the CMS ttbar analysis example
- Refactors how the systematics are looped over to make things more consistent by grouping systematics into lists of same-type corrections (e.g. event level vs. object level corrections) and making use of `coffea.analysis_tools.PackedSelection` to streamline the event selection
- Includes a new file called `corrections.json` which stores the correction/systematic shift information used by the `correctionlib` implementation
- Includes a new script called `make_corrections.py` which is a basic working example showing how the `corrections.json` file is created

Currently only implements the per-event systematics, but I would like to also include per-object systematics as well. The `make_corrections.py` script could also do with some annotating/descriptions. Lastly, the `.ipynb` file still needs to be synchronized with `ttbar_analysis_pipeline.py`

resolves #110


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -150,6 +141,7 @@ def __init__(self, disable_processing, io_branches):
         )
         self.disable_processing = disable_processing
         self.io_branches = io_branches
+        self.cset = correctionlib.CorrectionSet.from_file("corrections.json")
```
alexander-held, line 144: I did not think about this before: this will not work with Dask unless we also ship the file around, right? At the moment this would not get picked up. I imagine conceptually we could also build the `CorrectionSet` outside of the processor and hand it to `__init__` ready to go? Not sure if the pickling will automatically make that work though.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -182,112 +174,99 @@ def process(self, events):
...
-                    # nominal pT, but including 2-point systematics
-                    histogram.fill(
+                syst_var_name = f"{syst_var}"
+                # Break up the filling into event weight systematics and object variation systematics
+                if syst_var in event_systs:
+                    for i_dir,direction in enumerate(["up","down"]):
+                        # Should be an event weight systematic with an up/down variation
+                        if syst_var.startswith("btag_var"):
+                            i_jet = int(syst_var.rsplit("_",1)[-1])   # Kind of fragile
+                            wgt_variation = self.cset["event_systematics"].evaluate("btag_var",direction,region_jets.pt[:,i_jet])
```
alexander-held, line 252: One thing I'm wondering about here (not for this PR but maybe the future): by matching the name `btag_var` to the correction that is defined, the main thing here that differs per systematic variation (and means we need some dedicated per-systematic handling of the `.evaluate` call) is the information about which columns to feed in.

If we had an extra intermediate layer that knows which column the correction needs, and can automatically extract the right information, then the section here would perhaps not even need a dedicated treatment per systematic. In that case the `.evaluate` call equivalent would essentially get all the event information and then figure out internally what to pass on to `correctionlib`. I am not sure if this is all that feasible, or even a good idea. Maybe it only makes the code here look more streamlined by hiding the complication away?

cc @nsmith-

alexander-held (commented):
Thanks a lot, this is a great step forward. Two last comments and I think this is ready to go.

In the `ttbar_analysis_pipeline.py` file there are a lot of cases of no spaces before commas in function arguments. For consistency it would be good to have those spaces everywhere. We currently unfortunately do not yet use any formatting tool to make that automatic.

We also need to update the `.ipynb` version of the code, which can be achieved with `jupytext` via `jupytext --sync ttbar_analysis_pipeline.ipynb`. As the notebook outputs will also slightly change due to the pT variations being propagated properly, it would be good to -re-run the notebook subsequently to reflect that. I can also take care of that if you prefer.

ekauffma:
Commenting a quick correction for jupytext here.. the `jupytext --sync` command will make the python file match the notebook file, not the other way around. For the other way around, I believe you can temporarily delete the .ipynb file, run `jupytext --to notebook ttbar_analysis_pipeline.py` then re-run the notebook outputs, but I have not tried this...

alexander-held:
That's good to know, I thought both directions worked and it would just pick up the most recent changes. I did use `jupytext --sync analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py` and it looks like that worked successfully to sync the py  to ipynb.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -48,9 +48,11 @@
 
 import awkward as ak
 import cabinetry
+import correctionlib
 from coffea import processor
 from coffea.nanoevents import NanoAODSchema
-
+from coffea.analysis_tools import PackedSelection
+from servicex import ServiceXDataset
```
alexander-held, line 55:
```suggestion
```

We don't need that import here anymore, it moved to `utils`. This probably came in while updating against the changes in `main`.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -182,112 +174,99 @@ def process(self, events):
...
-            jet_filter = (events.Jet.pt * events[pt_var]) > 25 # pT > 25 GeV for jets (scaled by systematic variations)
-            selected_jets = events.Jet[jet_filter]
-
-            # single lepton requirement
-            event_filters = ((ak.count(selected_electrons.pt, axis=1) + ak.count(selected_muons.pt, axis=1)) == 1)
-            # at least four jets
-            pt_var_modifier = events[pt_var] if "res" not in pt_var else events[pt_var][jet_filter]
-            event_filters = event_filters & (ak.count(selected_jets.pt * pt_var_modifier, axis=1) >= 4)
-            # at least one b-tagged jet ("tag" means score above threshold)
+            # Note: This creates new objects, distinct from those in the 'events' object
```
alexander-held, line 200: Is this a feature of `awkward`, or how come this happens? It is not intuitive to me that the following lines create new objects.

Andrew42: I agree it isn't intuitive and it's something that occurs (afaik) for the `events` object, so I think it's more of a `coffea` thing than an `awkward` thing. I'll admit that I'm a bit leery of doing this since I don't fully understand/appreciate the details myself.

alexander-held: Seems like it is a cofffea thing, @nsmith- might have the details.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
Andrew42: Commenting here for documentation purposes. I think this does end up working with Dask, but only b/c the `TtbarAnalysis` class is defined within the jupyter notebook itself. If we instead had an `import` call to get the class, the Dask workers would indeed complain about not having the `corrections.json` file. I haven't explicitly tested this though.

alexander-held: Indeed, Dask looks to be working just fine with this current implementation.

alexander-held (approved):

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
nsmith-: It certainly could be done automatically, to the extent that the input names available from the correction object can be matched to columns in the file. That could be non-trivial, as it is only by convention.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
nsmith-: Fetching a sub-array from a record array in awkward does give you a shallow copy rather than a reference. e.g. adding or modifying the fields on the shallow copy will not mutate the original:
```python
array = ak.zip({"a": ak.zip({"x": np.ones(3), "y": np.ones(3)}), "b": np.ones(3)})
sub = array.a
sub["z"] = np.ones(3)
assert("z" not in array.a.fields)
```
Of course the underlying arrays are not copied, just the "dictionary" of references

alexander-held: Thanks a lot for confirming. Also very good to know that this will not copy the data around, which I had wondered about too. Then this should be behaving as we want.

PR #122, opened 2023-04-18
feat: add ML inference option to ttbar analysis pipeline

ekauffma:
An xgboost boosted decision tree is used to associate jets to their parent partons.


analyses/cms-open-data-ttbar/utils/__init__.py
```diff
@@ -1,12 +1,17 @@
...
 import hist
 import matplotlib as mpl
 import matplotlib.pyplot as plt
 import uproot
 from servicex import ServiceXDataset
 import numpy as np
+import awkward as ak
+
+with open("config.yaml") as config_file:
+    config = yaml.safe_load(config_file)
```
alexander-held, line 14: Could we instead pass in this config information via arguments to the new functions that use it? The main notebook has that config loaded anyway, and I think this will help avoid future issues where we may import functionality from this module from a place that may not have the config available.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -105,6 +106,113 @@
 with open("config.yaml") as config_file:
     config = yaml.safe_load(config_file)
 
+### ML-INFERENCE SETTINGS
+
+# enable ML inference
+USE_INFERENCE = True
+
+# enable inference using NVIDIA Triton server
+USE_TRITON = True
```
alexander-held, line 115: Should turn this off by default, will just fix in a separate PR as I need to submit another small fix unrelated to the changes in here anyway.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -241,6 +417,51 @@ def process(self, events):
                     # pick trijet candidate with largest pT and calculate mass of system
                     trijet_mass = trijet["p4"][ak.argmax(trijet.p4.pt, axis=1, keepdims=True)].mass
                     observable = ak.flatten(trijet_mass)
+                    
+                    if sum(region_selection)==0: continue
```
alexander-held, line 421: Are there any cases where no events pass the selection?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -405,25 +668,19 @@ def get_query(source: ObjectStream) -> ObjectStream:
 print(f"event rate per worker (pure processtime): {metrics['entries'] / metrics['processtime'] / 1_000:.2f} kHz")
 print(f"amount of data read: {metrics['bytesread']/1000**2:.2f} MB")  # likely buggy: https://github.com/CoffeaTeam/coffea/issues/717
 
-# %% [markdown]
-# ### Inspecting the produced histograms
-#
-# Let's have a look at the data we obtained.
-# We built histograms in two phase space regions, for multiple physics processes and systematic variations.
-
-# %%
```
alexander-held, line 414: I don't see a reason for why we can't keep this, will add it back.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -509,28 +777,53 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
-# ### What is next?
-#
-# Our next goals for this pipeline demonstration are:
-# - making this analysis even **more feature-complete**,
-# - **addressing performance bottlenecks** revealed by this demonstrator,
-# - **collaborating** with you!
-#
-# Please do not hesitate to get in touch if you would like to join the effort, or are interested in re-implementing (pieces of) the pipeline with different tools!
-#
-# Our mailing list is analysis-grand-challenge@iris-hep.org, sign up via the [Google group](https://groups.google.com/a/iris-hep.org/g/analysis-grand-challenge).
```
alexander-held, line 536: This cell should also stay. I'll add it back in.

alexander-held:
Changes going in via this:
```
* added ML inference via XGBoost BDT (with and without Triton)
* added calculation of input variables to BDT
* added visualization of variables derived from BDT
* updated event selection to match CMS Open Data workshop 2022
* expand func-adl query to provide additional variables needed
```

alexander-held (approved):

PR #125, opened 2023-04-20
fix: minor changes to docs

ekauffma:
- Fixed a few typos
- Title edits for readthedocs tab
- Removed "suggest edit" button (does not work with notebook symlinks and "open issue" button should be sufficient)
- Edited exclude pattern list in readthedocs configuration file to avoid warnings when building with sphinx
- Edited output of docs/event-plotting/plotEvents.ipynb
- Added badge for documentation in README


alexander-held (commented):
Could we please add a (non-badge) webpage link to the readme as well, perhaps next to the "we also have a dedicated webpage" link?

alexander-held (approved):
Thank you!

PR #126, opened 2023-04-26
fix: update I/O branches

ekauffma:
Includes script to calculate branches to use for IO and writes them to config file. Other notebook options are also moved to a config file.


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -172,44 +149,10 @@ def __init__(self, disable_processing, io_file_percent):
...
-                                         "Electron_miniPFRelIso_all","Electron_miniPFRelIso_chg","Electron_mvaFall17V2Iso",
-                                         "Electron_mvaFall17V2noIso","Electron_pfRelIso03_all","Electron_pfRelIso03_chg","Electron_r9",
-                                         "Electron_scEtOverPt","Electron_sieie","Electron_sip3d","Electron_mvaTTH","Electron_charge",
-                                         "Electron_cutBased","Electron_jetIdx","Electron_pdgId","Electron_photonIdx","Electron_tightCharge"])
-                
-            if self.io_file_percent not in [2.7, 4, 15, 25, 50]:
-                raise NotImplementedError("supported values for I/O percentage are 2.7, 4, 15, 25, 50")
-            
-            for branch in branches_to_read:
+            for branch in self.io_branches:
```
alexander-held, line 155: Just to be sure, does this works with Dask workers? I assume that `config["benchmarking"]["IO_BRANCHES"][config["benchmarking"]["IO_FILE_PERCENT"]]` will be passed by value into the processor so that's all good?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -445,7 +388,7 @@ def get_query(source: ObjectStream) -> ObjectStream:
 filemeta = run.preprocess(fileset, treename=treename)  # pre-processing
 
 t0 = time.monotonic()
-all_histograms, metrics = run(fileset, treename, processor_instance=TtbarAnalysis(DISABLE_PROCESSING, IO_FILE_PERCENT))  # processing
+all_histograms, metrics = run(fileset, treename, processor_instance=TtbarAnalysis(config["benchmarking"]["DISABLE_PROCESSING"], config["benchmarking"]["IO_BRANCHES"][config["benchmarking"]["IO_FILE_PERCENT"]]))  # processing
```
alexander-held, line 391: I'm guessing this will be too long for a cell, let's try to break these super long lines up to make them easier to read (also affects the notebook, scrolling is not super convenient). Perhaps break at the last comma. We're not enforcing strict formatting currently (should eventually probably just adopt `black`) but I've been trying to at least make lines not too super long.

analyses/cms-open-data-ttbar/config.json
alexander-held, line 1: This is going to be somewhat difficult to edit given it's all a line. Could we please have this made a bit more readable by saving with something like `json.dumps(config, sort_keys=True, indent=4)`? While this will auto-sort the keys, I would subsequently think that it might be best to have the simple and short settings all at the top, and then have the list of columns at the bottom of the file since users would not be expected to touch them.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ekauffma: Yes. I tested this.

analyses/cms-open-data-ttbar/config.json
```diff
@@ -0,0 +1,68 @@
+{
+    "global": {
+        "SERVICEX_IGNORE_CACHE": false,
```
alexander-held, line 3: I just noticed we're losing out on the comments in the notebook explaining these settings when doing this. JSON itself also doesn't natively support comments, so we should probably use YAML. Here's the YAML version:
```yaml
global:
  SERVICEX_IGNORE_CACHE: false
  AF: coffea_casa
benchmarking:
  CHUNKSIZE: 200000
  AF_NAME: coffea_casa
  SYSTEMATICS: all
  CORES_PER_WORKER: 2
  NUM_CORES: 4
  DISABLE_PROCESSING: false
  IO_FILE_PERCENT: 4.1
  IO_BRANCHES:
    4.1:
      - Jet_pt
      - Jet_eta
      - Jet_phi
      - Jet_btagCSVV2
      - Jet_mass
      - Muon_pt
      - Electron_pt
    15:
      - LHEPdfWeight
      - GenPart_pdgId
      - CorrT1METJet_phi
    25:
      - LHEPdfWeight
      - GenPart_pt
      - GenPart_eta
      - GenPart_pdgId
      - LHEScaleWeight
    50:
      - LHEPdfWeight
      - GenPart_pt
      - GenPart_eta
      - GenPart_phi
      - GenPart_pdgId
      - GenPart_genPartIdxMother
      - GenPart_statusFlags
      - GenPart_mass
      - LHEScaleWeight
      - GenJet_pt
      - GenPart_status
      - LHEPart_eta
      - LHEPart_phi
      - LHEPart_pt
      - GenJet_eta
      - GenJet_phi
      - Jet_eta
      - Jet_phi
      - SoftActivityJet_pt
      - SoftActivityJet_phi
      - SoftActivityJet_eta
      - GenJet_mass
      - Jet_pt
      - Jet_mass
      - LHEPart_mass
      - Jet_qgl
      - Jet_muonSubtrFactor
      - Jet_puIdDisc
```
This switches the strings for the I/O fraction to numbers, can turn them into strings again if this is any problem.

alexander-held: This allows for including the original comments again like
```yaml
global:
  SERVICEX_IGNORE_CACHE: false  # ServiceX: ignore cache with repeated queries
```

analyses/cms-open-data-ttbar/config.yaml
```diff
@@ -0,0 +1,82 @@
...
+  # ServiceX: ignore cache with repeated queries
+  SERVICEX_IGNORE_CACHE: false
+  
+  # analysis facility: set to "coffea_casa" for coffea-casa environments, "EAF" for FNAL, "local" for local setups
+  AF: coffea_casa
+  
+benchmarking:
+
+  # chunk size to use
+  CHUNKSIZE: 200000
```
alexander-held, line 12: The decrease is only needed for the ML part, right? Here we're still fine with 500k?

alexander-held (approved):
Looks all good, thank you very much!

PR #127, opened 2023-04-27
feat: cms ttbar training notebook

ekauffma:
Add notebook to train BDT (assigns jets to their parent partons).
The models trained in this notebook can be used in the ttbar_analysis_pipeline.ipynb notebook.


analyses/cms-open-data-ttbar/jetassignment_training.py
```diff
@@ -0,0 +1,1347 @@
...
+    return processor.column_accumulator(np.array(a))
+
+processor_base = processor.ProcessorABC
+class JetClassifier(processor_base):
+    def __init__(self, permutations_dict, labels_dict):
+        super().__init__()
+        self.permutations_dict = permutations_dict
+        self.labels_dict = labels_dict
+    
+    def process(self, events):
```
alexander-held, line 317: Perhaps we do not have much of an alternative for now, but duplicating the processor code (and a few other piecse) from the main notebook is going to make maintenance difficult. Do you think that conceptually it would be possible to avoid all duplication and import common pieces from a common dependency? We do not need to do that here, but it would be interesting to know if that might in principle work (and we can refactor later on).

ekauffma: I can probably remove the trijet combination calculation from this notebook. I just left it in so that we could compare the two methods for top mass reconstruction. Do you think it is best if this is removed then?

ekauffma: I think that's the only part of the processor code which is really duplicated...

alexander-held: Some of the configuration options at the top of the notebook also are, the metadata handling below for normalization is the same, and some cuts are as well. They aren't huge overlaps but I think what we need to eventually move towards is a setup where all this functionality is packaged up in common utility functions that are called from any processor we use for this analysis.

I would say leave the trijet combination in here for now and in the longer term we can try to find a solution that avoids overlaps.

ekauffma: Yeah.. I think without dependency management for dask workers, handling this in a great way is not possible at the moment

alexander-held: Could you please open an issue noting down these things here so we don't forget about them and can follow up on them in the future? We can keep this PR here the way it is.

ekauffma: #129

analyses/cms-open-data-ttbar/utils/__init__.py
```diff
@@ -153,3 +194,386 @@ def get_data_rootfiles_uri(self, query, as_signed_url=True, title="Untitled"):
...
+        
+        fig.show()
+
+        figs.append({"figure": fig, "region": channel_name})
+    
+    
+    return figs
+    
+# modified from generate_config in https://github.com/triton-inference-server/fil_backend/blob/main/qa/L0_e2e/generate_example_model.py
+def generate_triton_config(model_name,
```
alexander-held, line 515: Is this used anywhere?

alexander-held: Missed it, this is used in the training notebook (I was wondering whether it is used for inference, but it is written out and then later used).

analyses/cms-open-data-ttbar/jetassignment_training.py
```diff
@@ -0,0 +1,1393 @@
...
+#
+#
+# <mark>top<sub>lepton</sub></mark> and <mark>top<sub>hadron</sub></mark> jets do not necessarily correspond to top/antitop, respectively. The <mark>top<sub>lepton</sub></mark> jet is defined as having a lepton/neutrino pair as cousins, where the <mark>top<sub>hadron</sub></mark> jet is defined as having two jets as cousins. The <mark>W</mark> jets are not distinguished from each other.
+#
+# The strategy for solving this problem is to train a boosted decision tree to find the correct assignments for each jet. Since we consider four jets per event with three unique labels (<mark>W</mark>, <mark>top<sub>lepton</sub></mark>, and <mark>top<sub>hadron</sub></mark>), there are twelve possible combinations of assignments:
+#
+# <img src="utils/jetcombinations.png" alt="jetcombinations" width="700"/>
+#
+# The combination with the highest BDT score will be selected for each event.
+# ____
```
alexander-held, line 31: This doesn't seem to render as line, at least not on GitHub.

analyses/cms-open-data-ttbar/jetassignment_training.py
```diff
@@ -0,0 +1,1393 @@
...
+
+# %% tags=[]
+fileset = utils.construct_fileset(N_FILES_MAX_PER_SAMPLE, 
+                                  use_xcache=False)
+
+# get rid of everything except ttbar__nominal for training purposes
+fileset_keys = list(fileset.keys())
+for key in fileset_keys:
+    if key!="ttbar__nominal":
+        fileset.pop(key)
```
alexander-held, line 422: a bit more compact suggestion:
```python
fileset = {"ttbar__nominal": fileset["ttbar__nominal"]}
```

analyses/cms-open-data-ttbar/jetassignment_training.py
```diff
@@ -0,0 +1,1393 @@
...
+run = processor.Runner(executor=executor, schema=NanoAODSchema, savemetrics=True, metadata_cache={}, 
+                       chunksize=CHUNKSIZE)
+
+# preprocess
+filemeta = run.preprocess(fileset, treename="Events")
+
+# process
+output, metrics = run(fileset, 
+                      "Events", 
+                      processor_instance = JetClassifier(permutations_dict, labels_dict))
```
alexander-held, line 444: It would be probably useful to save the output after this step, allowing the subsequent training to be re-run from just the saved output. Having to re-run `coffea` every time might not be super convenient. We can do that in a follow-up though.

analyses/cms-open-data-ttbar/jetassignment_training.py
alexander-held: I think it needs a blank line above and below to work.

alexander-held:
```suggestion
#
# ____
#
```

ekauffma:
> It looks like something went wrong here and the diff now shows a ton of files as changed, any idea why?

I have no idea. I did not see this when I made the pull request.. I'll take some time to look into this.

analyses/cms-open-data-ttbar/jetassignment_training.py
```diff
@@ -0,0 +1,842 @@
...
+# ```
+# mc alias set triton http://$BUCKET_HOST $AWS_ACCESS_KEY_ID $AWS_SECRET_ACCESS_KEY
+# mc cp -r reconstruction_bdt_xgb triton/$BUCKET_NAME/reconstruction_bdt_xgb/
+# ```
+#
+# The server may need to be restarted in order to load the model.
+
+# %% tags=[]
+# remove model directory after uploading to triton
+# !rm -r reconstruction_bdt_xgb
```
alexander-held, line 684: It would be good to remove this feature by default to allow the model files to be still available after running the notebook. Perhaps we can delete the folder before creating new files to avoid any kind of collisions, but even that is maybe optional.

alexander-held:
Please also have a look at the config written out, there are some issues with brackets not being closed.

PR #139, opened 2023-05-17
fix: do not import tritonclient.grpc unless needed

eguiraud:
The import can error out if triton is not set up correctly, but we do not care whether triton is set up correctly unless we actually want to use it.

Given the error this could also just be a missing dependency, but I think a delayed import makes sense in general.

The error I got after creating an environment according to requirements.txt:

```
   Traceback (most recent call last):
     File "/home/blue/Tools/miniconda3/envs/agc-py311/lib/python3.11/site-packages/tritonclient/grpc/__init__.py", line 30, in <module>
       from google.protobuf.json_format import MessageToJson
   ModuleNotFoundError: No module named 'google.protobuf'

   The above exception was the direct cause of the following exception:

   Traceback (most recent call last):
     File "/home/blue/Scratchpad/work/agc/analysis-grand-challenge/analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py", line 63, in <module>
       import tritonclient.grpc as grpcclient
     File "/home/blue/Tools/miniconda3/envs/agc-py311/lib/python3.11/site-packages/tritonclient/grpc/__init__.py", line 39, in <module>
       raise RuntimeError(
   RuntimeError: The installation does not include grpc support. Specify 'grpc' or 'all' while installing the tritonclient package to include the support
```


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
alexander-held, line 1: We'll also need to sync this to the notebook file.

ekauffma: I should probably add jupytext to the CI

eguiraud: for this PR I can just do it manually, I'll amend the commit

eguiraud: done

PR #141, opened 2023-05-17
Add a conda environment file that works for Python 3.11

eguiraud:
The current requirements.txt does not work for 3.11 because of the specified versions of llvmlite and numba.

This might not be the _best_ env for Python 3.11, that is, most packages could probably be upgraded to newer versions, but at least this env runs.


matthewfeickert (commented):
Why is this a conda environment file and not just a requirements.txt as you're installing everything from PyPI? As most of these packages are available on conda-forge is there a reason to not install them from there and then only install any remainders from PyPI?

alexander-held:
The idea behind that was fixing the Python version as well, since with the previous file the switch in Python versions caused issues.

I think we should have a single list to maintain though, so if we switch to this, then also drop the old requirements.txt.

matthewfeickert:
> I think we should have a single list to maintain though, so if we switch to this, then also drop the old requirements.txt.

If you switch to using lock files then this becomes much easier to even maintain multiple high level environment files and then let the locking system solve the full environment. c.f. https://github.com/iris-hep/analysis-systems-base

eguiraud:
This PR is basically just to document that 3.11 does not work with the current `requirements.txt` and that it does with _something like_ this conda env. I agree this is still not the best setup :smile: 

@matthewfeickert did you mean to say noxfile rather than lock file? or: what is a python lock file?

eguiraud:
aaah found it, https://github.com/conda/conda-lock

eguiraud:
For future reference these are the packages that are not available via conda and conda-forge:

```
fastjsonschema
func-adl
func-adl-servicex
ipython-genutils
jupyter-console
jupyter-events
jupyterlab-pygments
jupyterlab-widgets
make-it-sync
mplhep-data
msgpack
notebook_shim
prometheus-client
pure-eval
qastle
servicex
stack-data
```

eguiraud:
This is really more of a problem statement than a PR, so I created #144 to take the place of this PR.

PR #146, opened 2023-05-26
fix: imports, remove unused variable

eguiraud:


analyses/cms-open-data-ttbar/utils/__init__.py
```diff
@@ -1,5 +1,5 @@
-import asyncio
 import json
+import logging as log
```
alexander-held, line 2: I see this is a bug in the current implementation, but could we just fix the module use and make that `logging` instead of changing the name here? I'd prefer to not rename if not needed.

eguiraud: sure, done!

eguiraud:
(btw a linter in the CI should catch these things)

alexander-held:
Indeed, our CI is very much in need of some improvements. `flake8` / `ruff` would have caught this. We just need to set up some reasonable configuration for them as we're not PEP8 compliant in some regards, especially line length (but that can be fixed).

alexander-held (approved):
Looks all good to me, thanks! I'll go ahead with this as the `results_odd` fix seems reasonable. @ekauffma, if anything about this doesn't make sense we can fix it in a follow-up.

PR #148, opened 2023-05-26
fix: ruff linter errors

eguiraud:
@alexander-held there is only one error remaining (besides errors due to lines that are too long):

```
workshops/agctools2022/statistical-inference/exclusion/src/exclusion/interpolate.py:187:57: F821 Undefined name `constraintsDict`
Found 1 error.
```

Do you have any suggestion how to solve that one?

This is in preparation to [the next PR](https://github.com/iris-hep/analysis-grand-challenge/pull/154) which adds a ruff check to the CI.


alexander-held:
I'm not sure who wrote that revised version of `interpolate.py`, perhaps it was @kratsg or @matthewfeickert?
This line specifically is the problem:

https://github.com/iris-hep/analysis-grand-challenge/blob/859e3e0d6dd6529684086d3397b90d06f00f3682/workshops/agctools2022/statistical-inference/exclusion/src/exclusion/interpolate.py#L187

`constraintsDict` is never defined. In the HistFitter original version it is: https://github.com/histfitter/histfitter/blob/1492543dcb4df8364f9f6ef8e6d044d8a968db37/scripts/harvestToContours.py#L338-L347.

workshops/agctools2022/statistical-inference/exclusion/src/exclusion/visualize.py
```diff
@@ -125,7 +125,3 @@ def plot_contour(ax, results, **kwargs):
         )
 
     return ax
-
-
-if __name__ == "__main__":
-    main()
```
alexander-held, line 131: I guess this gets removed because `main` isn't defined? Not sure either what the purpose of it was. @kratsg / @matthewfeickert do you know?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -678,13 +679,13 @@ def get_query(source: ObjectStream) -> ObjectStream:
 all_histograms["hist"][120j::hist.rebin(2), "4j1b", :, "nominal"].stack("process")[::-1].plot(stack=True, histtype="fill", linewidth=1, edgecolor="grey")
 plt.legend(frameon=False)
 plt.title("$\geq$ 4 jets, 1 b-tag")
-plt.xlabel("$H_T$ [GeV]");
+plt.xlabel("$H_T$ [GeV]")
```
alexander-held, line 682: These semicolons all have the purpose to suppress output that would otherwise be printed as far as I remember. Maybe there is a better way (we could assign `_ = ...` to achieve the same), but if not then I'd say it is better to keep them and disable the associated warning code.

kratsg:
> I'm not sure who wrote that revised version of interpolate.py, perhaps it was @kratsg or @matthewfeickert?

I suspect @lukasheinrich was the one that did the modifications... unless it was @matthewfeickert . I didn't do this.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
eguiraud: I see, another quirk of syncing the `.py` and the `.ipynb` versions I guess. I will just silence this type of linting error.

eguiraud: Done

alexander-held:
Hi @lukasheinrich, @matthewfeickert, do either of you remember this exclusion contours script and can help us identify how to fix it?

eguiraud:
@matthewfeickert ping :smile: (I realize there are still 20 minutes before 24 hours have passed, but I was doing a pass over my open PRs now sorry!)

alexander-held:
Hi @matthewfeickert, friendly ping here. If we cannot figure out what this was let's just go ahead and not touch those files.

matthewfeickert:
I'll look at this after lunch, but this indeed was me as this is what I threw together for a demo (maybe for PyHEP 2022?) which is https://pypi.org/project/exclusion/ (the source code currently lives on the `feat/add-cookie` branch of https://github.com/matthewfeickert/exclusion as I guess I never even finished that loop). Some of this stuff got ported from original workflows in HistFitter and as I was rushing I didn't clean some of the patterns that exist there. So I'll look at this and try to clean things, but I think the long and the short of it is that basically anything `ruff` wants to do should go in, unless you want to flip this out for just using `exclusion` from PyPI and then someone pushing me to maintain it (could probably find some interest in ATLAS SUSY groups I would hope).

matthewfeickert (changes requested):
> ```
> workshops/agctools2022/statistical-inference/exclusion/src/exclusion/interpolate.py:187:57: F821 Undefined name `constraintsDict`
> Found 1 error.
> ```
> 
> Do you have any suggestion how to solve that one?

Given that this is undefined code that got slap-dash copied from other scripts over (not good on many levels including potentially licensing), as @alexander-held mentioned about it only appearing once

```console
$ git rev-parse HEAD
54b43812a4219748af82b86cad965e08d72796cb
$ git grep "constraintsDict"
workshops/agctools2022/statistical-inference/exclusion/src/exclusion/interpolate.py:                for (constrainedVar, constrainedVal) in constraintsDict.iteritems()
```

I'd just say remove the offending chunks as they obviously aren't getting used for any plotting that was happening.

alexander-held (approved):

PR #149, opened 2023-05-29
feat: add utility tool to validate histogram contents against a reference

### LINKED ISSUE #136: Add a way to validate output histograms against a trusted reference
In first approximation it would be enough to publish reference histogram content, in whatever easily readable format (even JSON, or serialized dictionaries of numpy arrays).

A simple CLI tool or similar that runs the comparison between two such JSON files could be provided on top. 

Complications:
- one systematic variation depends on RNG numbers, so bin values won't be stable (especially with multi-thread/out-of-order execution): I think it's ok to clearly mark it as such and only expect a very approximate equality there
- bin values depend on how many files per process are used: I think we should provide values for at least 1 file per process (for quick validation) and for the full dataset (for full validation)

eguiraud:
A reference file for the configuration with N_FILES_MAX_PER_SAMPLE=1 is provided as well.

The tool:
- compares histogram edges and contents against a JSON reference file
- can also dump the contents of the histograms as json in order to produce new reference files
- ignores mismatches in the contents (but not the edges) of histograms with name containing "pt_res_up", because the content of these histograms depends on a random number generator
- only considers the highest namecycle in case there are multiple cycles for the same histogram (e.g. `name;1` and `name;2`)

Fixes #136 .


eguiraud:
That would be just `python validate_histograms.py --dump-json` (assuming the input has the default file name `histograms.root`, otherwise you also pass that to the `--histos` option).

Is that what you mean? I'm not sure where to put that other than the `--help` as it's now?

alexander-held:
Perfect, I hadn't looked closely enough at the code yet. No need to change anything then!

ekauffma:
Can we add a description of this tool to the README for cms-open-data-ttbar?

eguiraud:
@ekauffma sure, done!

analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -0,0 +1,74 @@
...
+        cycle = int(cycle[0]) if len(cycle) > 0 else -1
+        if name in histos and cycles["cycle"] > cycle:
+            continue # found a lower cycle for a histogram we already have
+
+        cycles[name] = cycle
+        histos[name]["edges"] = v.axis().edges().tolist()
+        histos[name]["contents"] = v.counts(flow=True).tolist()
+    return histos
+
+def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
```
alexander-held, line 38: This requires Python 3.9+, so I would suggest to add an `from __future__ import annotations` for backwards compatibility at the beginning of the file. While [NEP 29](https://numpy.org/neps/nep-0029-deprecation_policy.html) puts us already in the 3.9+ age, I imagine not everyone is up-to-date with that yet.

analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -0,0 +1,74 @@
...
+def as_dict(f: uproot.ReadOnlyDirectory) -> dict[str, dict]:
+    histos = defaultdict(dict)
+    cycles: dict(str, int) = {}
+    for k, v in f.items():
+        assert isinstance(v, uproot.behaviors.TH1.Histogram)
+        # this assumes that the rightmost ";" (if any) comes before a namecycle
+        name, *cycle = k.rsplit(";", 1)
+        cycle = int(cycle[0]) if len(cycle) > 0 else -1
+        if name in histos and cycles["cycle"] > cycle:
+            continue # found a lower cycle for a histogram we already have
```
alexander-held, line 31: I think `uproot` would by default just use the highest/latest cycle, so we might not need that, but not 100% sure (cc @jpivarski). On the other hand, I cannot think spontaneously of any harm caused by it either.

analyses/cms-open-data-ttbar/validate_histograms.py
alexander-held: I agree for this repo, but am hoping that this might be useful more broadly and the cost of the extra line is negligible for this case.

analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -0,0 +1,74 @@
...
+
+    return errors
+
+if __name__ == "__main__":
+    args = parse_args()
+    with uproot.open(args.histos) as f:
+        histos = as_dict(f)
+
+    if args.dump_json:
+        print(json.dumps(histos))
```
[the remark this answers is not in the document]
eguiraud: The reference file is formatted via `jq`, e.g. by calling `python validate_histos.py --dump-json | jq > reference.json`.

I changed the invocation here to `json.dumps(histos, indent=2)` which does something pretty similar.

analyses/cms-open-data-ttbar/validate_histograms.py
eguiraud: At least when looping over keys, uproot exposes all namecycles:

```python
In [2]: f = uproot.open("f.root")

In [3]: f.keys()
Out[3]: ['h;2', 'h;1']

In [4]: dict(f.items())
Out[4]:
{'h;2': <TH1D (version 3) at 0x7f5730bb7950>,
 'h;1': <TH1D (version 3) at 0x7f56df9d3fd0>}
```

The removal of the `;N` part of the histogram name is probably desirable anyways.

alexander-held: Right, I'm not aware of a way to get this list without the cycles included in the names. For simplicity maybe easiest to just strip the cycle name and then proceed by lookup without cycle names?

analyses/cms-open-data-ttbar/validate_histograms.py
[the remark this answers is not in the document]
eguiraud: I updated the reference

analyses/cms-open-data-ttbar/validate_histograms.py
eguiraud: that requires one loop to get all names (stripping the cycles as they come in) and then another loop to redo the lookup into the file.

nothing here is performance critical so I can do that if you want?

alexander-held: I'm not worried about performance, it just seemed to me that we can simplify this:
```python
def as_dict(f: uproot.ReadOnlyDirectory) -> dict[str, dict]:
    histos = defaultdict(dict)
    for k, v in f.items():
        assert isinstance(v, uproot.behaviors.TH1.Histogram)
        # this assumes that the rightmost ";" (if any) comes before a namecycle
        name = k.rsplit(";", 1)[0]
        histos[name]["edges"] = v.axis().edges().tolist()
        histos[name]["contents"] = v.counts(flow=True).tolist()
    return histos
```
Given that `uproot` will automatically pick the latest cycle, I think this is equivalent and allows to drop the manual handling of it?

eguiraud: ah, no that's not equivalent, `for k, v in f.items()` loops over _all_ cycles and you'll end up with the contents of _one of those_ in `histos[name]` (not sure which, I guess either the highest or the lowest cycle, but I'm not sure it's guaranteed to be stable) at the end of the loop.

Only accessing a key via `f["histo_name"]` returns the highest cycle.

What I meant above is that accessing the uproot keys twice instead of once we can write this:

```python
def as_dict(f: uproot.ReadOnlyDirectory) -> dict[str, dict]:
    histos = defaultdict(dict)
    # this assumes that the rightmost ";" (if any) comes before a namecycle
    names = set(k.rsplit(";", 1)[0] for k in f)
    for name in names:
        h = f[name]
        assert isinstance(h, uproot.behaviors.TH1.Histogram)
        histos[name]["edges"] = h.axis().edges().tolist()
        histos[name]["contents"] = h.counts(flow=True).tolist()
    return histos
```

and now that I see it written down I do like it better than the current code :smile: 

What do you think?

alexander-held: You're right of course, and yes I like that latest version too. Let's use that and get this merged. :)

eguiraud: done

alexander-held (approved):

Issue #130, opened 2023-05-02
Follow-up items to ML extension of analysis


alexander-held:
Collecting follow-up items to #122 here that are not crucial to be addressed immediately in that PR but can be revisited. cc @ekauffma 

- [ ] understand large change in event yields with new cuts (almost an order of magnitude less events), though the new yields are consistent with what CMS had for the 2022  open data workshop in https://cms-opendata-workshop.github.io/workshop2022-lesson-ttbarljetsanalysis/02-coffea-analysis/index.html#plotting
- [ ] investigate possibility of merging histogram-writing code into single function that avoids hardcoding information where possible
- [ ] harmonize object names (includes also the ML training notebook and documentation probably) from e.g. "top_hadron jet" to  "b_{had top}" etc, focusing the names of the b-tagged jet on "b" instead of "top"

alexander-held:
The current implementation has been stable for a long time, closing this as resolved in preparation for tagging v2.

PR #153, opened 2023-06-07
feat: improve integration of ml task

ekauffma:
This PR partially addresses issue #130 with the following features:

- fix: get rid of unnecessary particle dependency
- first plots of ML features are turned into a grid of plots to save space
- `USE_INFERENCE` flag required for ML model fitting
- `get_query` method updated to accommodate new cuts and selectively provide branches needed for ML features depending on flag `USE_INFERENCE`


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -528,30 +530,81 @@ def postprocess(self, accumulator):
...
+                                       "Jet_phi": h.Jet_phi,
+                                       "Jet_qgl": h.Jet_qgl,
+                                       "Jet_btagCSVV2": h.Jet_btagCSVV2,
+                                       "Jet_jetId": h.Jet_jetId,
+                                       "event": h.event,
+                                      })
+    if USE_INFERENCE:
+        return selection
+    
+    # some branches are only needed if USE_INFERENCE is turned on
```
alexander-held, line 591:
```suggestion
    # some branches are not needed if USE_INFERENCE is turned off
```

This is really minor, but as the next line describes the "off" case, I think this is clearer.

alexander-held (commented):
Thanks for taking care of some of these follow-up items! Does the new query pass the histogram comparison script?

ekauffma:
> 

I am not sure... the query takes forever to run, but I will go ahead and test this

PR #154, opened 2023-06-07
ci: add ruff linter check

eguiraud:
Depends on #148 , but since the only pending item in #148 is about a script in the `workshop` directory and we exclude that directory from the ruff check, #148 could also be merged as it is and it would make the linter check from this PR green.

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,11 +4,16 @@ on:
   push:
     branches: [ main ]
   pull_request:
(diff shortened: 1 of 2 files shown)


eguiraud:
Rebased on main now that #148 is merged.

eguiraud:
Remaining failures are fixed by #155 .

alexander-held (approved):
Great to see this passing now, thank you!

PR #157, opened 2023-06-12
feat: provide reference files for various numbers of files processed

ekauffma:
The previous reference json for 1 file/sample was generated before the `scale_var` fix in #152 . I re-ran `validate_histograms.py` to update this reference file.


alexander-held (commented):
I didn't think about the consequence of moving to #149 (cc @eguiraud): this makes the order of keys in `histos` random and consequently means the keys can move around. Let's address this also in this PR: by adding `sort_keys=True` to `json.dumps` inside `validate_histograms.py` we will keep the same order across executions and limit the amount of changes if we need similar types of updates in the future.

alexander-held (approved):
Looks good to me, thank you!

PR #158, opened 2023-06-12
feat: dependency management with cloudpickle

ekauffma:
`cloudpickle` seems to be a good solution for dependency management in the ttbar notebook, so it is adopted in this PR. `utils` is split into different modules and many boilerplate methods and machine learning task details are moved outside of the main notebook. Using `cloudpickle.register_pickle_by_value(utils)`, the `dask` workers have access to `utils`. This PR also addresses items 2 and 5 in #130


ekauffma:
Note: tested this with all possible combinations of `USE_DASK`, `USE_SERVICEX`, `USE_INFERENCE`, and `USE_TRITON`. Everything runs (at least for 1 file case)

ekauffma:
As requested by @alexander-held, these are the changes that are a bit more than just code-rearranging:

- utils/file_output.py and ttbar_analysis_pipeline.py lines 127-138
- ttbar_analysis_pipeline.py lines 681-684
- utils/ml.py lines 10-17

I believe this is all, but let me know if anything else feels like a substantial change!

analyses/cms-open-data-ttbar/README.md
```diff
@@ -7,7 +7,6 @@ This directory is focused on running the CMS Open Data $t\bar{t}$ analysis throu
 | GetIOBranches.py              | This script uses `nanoaod_branch_ratios.json` to calculate a list of branches to read in specific fractions of data for I/O benchmarking tests. |
 | cabinetry_config.yml          | This is a config file for `cabinetry`, which is used to build a statistical workspace.                                                          |
 | cabinetry_config_ml.yml       | This is another `cabinetry` config file, but specifically for the Machine Learning observables.                                                 |
-| config.yml                    | This is a general config file to handle different options for running the analysis.                                                             |
```
alexander-held, line 10: Let's add the .py version of the config here instead.

alexander-held:
Note: this also adopts a python-based config (former YAML).

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -219,92 +114,74 @@ def get_features(jets, electrons, muons, permutations_dict):
...
+                                  utils.config["global"]["BIN_LOW"], 
+                                  utils.config["global"]["BIN_HIGH"], 
+                                  name="observable", 
+                                  label="observable [GeV]")
+                .StrCat([], name="process", label="Process", growth=True)
+                .StrCat([], name="variation", label="Systematic variation", growth=True)
+                .Weight()
+            )
+        
         self.use_dask = use_dask
```
alexander-held, line 140: We don't seem to be using this anymore as far as I can see? It was previously controlling the model setup for Triton, how is that working now? Can we remove this property here?

analyses/cms-open-data-ttbar/utils/client.py
alexander-held, line 1: Can we name this `clients.py` instead as it deals with multiple different ones please (leaving open the possibility to add more in the future too)?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -845,22 +678,9 @@ def get_query(source: ObjectStream) -> ObjectStream:
...
-# We can see very good post-fit agreement.
-
-# %% tags=[]
 model_prediction_postfit = cabinetry.model_utils.prediction(model, fit_results=fit_results)
-figs = cabinetry.visualize.data_mc(model_prediction_postfit, data, close_figure=True, config=config)
-figs[0]["figure"]
-
-# %% tags=[]
-figs[1]["figure"]
+figs = cabinetry.visualize.data_mc(model_prediction, data, close_figure=True, config=config)
```
alexander-held, line 682: This now effectively only does the saving to file, right? Since the line below does the actual visualization, it does not save too much of a purpose anymore. Let's add a comment to the next line then saying that this re-implements the visualization in a grid view (hopefully we can eventually replace that, but there's an issue already to track this anyway).

alexander-held (approved):
Looks all good. Thank you!

Issue #162, opened 2023-06-22
Path towards v1.1


alexander-held:
Prepare a v1.1 tag, which will correspond to the v1 analysis task. The task will **not** change. We presumably will branch off of v1.0.0 for this.

cc @eguiraud for related discussions

- [x] backport #149 and provide updated histogram references
  - done via #170
- [x] change histogramming: save histograms in binning as provided by coffea output, subsequently re-bin via cabinetry (https://github.com/scikit-hep/cabinetry/issues/412)
  - we declare this as "not affecting analysis task" as the binning used in the histograms is not explicitly specified, just implicitly starts to look like this now due to the functionality of the histogram validation
  - done via #165
- [x] port a fix for #131 which affects v1.0.0, so this is currently a (known) bug in that implementation that has since been fixed via `correctionlib` (we probably want some simpler one-off fix, as there will be no additional development on v1.* following this presumably)
  - done via https://github.com/iris-hep/analysis-grand-challenge/pull/163
- [x] fix cabinetry binning labels in config (this also should go into `main` after #158)
  - done for v1 via #165

Moving forward: major versions are distinct analysis tasks, minor/patch reserved for implementations.

alexander-held:
We should also update the lower bound when re-binning, as that currently does not align with existing bin edges
```python
all_histograms[120j::hist.rebin(2), "4j2b", "ttbar", "nominal"]
```
This should be equivalent to `110j` (which is a bin boundary).

alexander-held:
v1.1 is now released: https://github.com/iris-hep/analysis-grand-challenge/releases/tag/v1.1.0

PR #163, opened 2023-06-23
fix: pt variation change carried through signal region calculations

ekauffma:
Addresses 3rd point in #162 (pt variations are carried through by overwriting jet object)


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -236,6 +236,22 @@ def process(self, events):
                 elif region == "4j2b":
                     region_filter = ak.sum(selected_jets.btagCSVV2 > B_TAG_THRESHOLD, axis=1) >= 2
                     selected_jets_region = selected_jets[region_filter]
+                    
+                    pt_var_modifier = (
+                        events[event_filters][region_filter][pt_var]
```
alexander-held, line 241:
```suggestion
                        events[pt_var][event_filters][region_filter]
```

I think this does the same and is more intuitive to read, at least for me: we attached `pt_var` directly to the `event` previously. This restores the consistency with the `else` version two lines below as well. Since the 4j1b version is the same, please change it there as well.

alexander-held (commented):
As a sanity check, could you please compare output histograms to https://github.com/eguiraud/analysis-grand-challenge-coffea/tree/correct-outputs (single file should be enough)? That one is built from the correctionlib version with back-ported W+jets scale variation fix.

ekauffma:
This version seems to have the new object cuts. Is there a version equivalent to v1?

ekauffma:
Documented discrepancies here:
[compare_result (1).txt](https://github.com/iris-hep/analysis-grand-challenge/files/11870888/compare_result.1.txt)

Seems to be some discrepancy due to floating point error.

alexander-held:
That's interesting, looking at the first example
```
got      [43415.260890439, 90524.50293484158, 124669.85937904267, 78590.96671041285, 51739.84456109777, 36289.230414965794, 25796.08326822223, 20466.19299444227, 15611.67260972688, 12753.150585139763, 10134.699515096338, 8746.319925277914, 54751.576864877905]
expected [43415.260890439, 90524.50293484157, 124668.95084443182, 78590.93441945997, 51740.78538666148, 36289.230414965794, 25796.08326822223, 20466.19299444227, 15611.67260972688, 12753.150585139763, 10134.699515096338, 8746.319925277914, 54751.576864877905]
```
it looks like there are just some events migrating between bins. For example, these three bins
```
got 124669.85937904267, 78590.96671041285, 51739.84456109777
expected 124668.95084443182, 78590.93441945997, 51740.78538666148
```
are presumably caused by two events moving between the outer and inner bin. From a quick look at other cases, that looks like it generally explains the observed behavior. It should be fine to move forward with this then. Thanks for checking!

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -236,6 +236,22 @@ def process(self, events):
...
+                        {"pt": selected_jets_region.pt * pt_var_modifier,
+                         "eta": selected_jets_region.eta,
+                         "phi": selected_jets_region.phi,
+                         "mass": selected_jets_region.mass,
+                         "btagCSVV2": selected_jets_region.btagCSVV2},
+                        with_name="PtEtaPhiMLorentzVector"
+                    )
 
                     # reconstruct hadronic top as bjj system with largest pT
                     # the jet energy scale / resolution effect is not propagated to this observable at the moment
```
alexander-held, line 257: This line needs to be removed, as this update takes care of propagating things correctly.

alexander-held:
I've looked into this a bit, code for reference is in https://gist.github.com/alexander-held/d44d089b7a71f25bae17a03665325400. This compares the `correctionlib`-based version from @eguiraud's branch to this PR (without `correctionlib`). I'm essentially only looking at nominal ttbar, single file, and a zoomed in histogram with bin edges `[189.999, 190.0, 190.001]`. I am printing out values of the 4j2b observable that falls within that window with 190 GeV subtracted from it.

**this PR:**
```
-0.0002548726
0.0003070348
0.0001955073
0.0003590095
```
resulting in bin yields of  `[3.69506563 3.69506563]`

`correctionlib` version:
```
0.0000000000
0.0000000000
0.0000000000
0.0000000000
```
resulting in bin yields of `[1.84753282 5.54259845]`

The differences are caused by events migrating across the boundary. What is not clear to me is why all the events in the `correctionlib` version seem to be ending up at exactly 190 GeV.

alexander-held:
Looking closer into types, the `correctionlib` version ends up with `observable` being
```
<Array [517, 229, 192, ... 262, 3.84e+03, 198] type='48881 * ?float32'>
```
while this PR has
```
<Array [367, 245, 383, 627, ... 352, 236, 617] type='48939 * ?float64'>
```
I guess the difference here comes down to 32 vs 64bit floats, not sure what exactly causes the difference in setups.

alexander-held (approved):
Looks good, thank you!

PR #164, opened 2023-06-27
Coffea2023 - minimal working notebook

ekauffma:
created version of ttbar notebook that processes one file (ttbar) and produces histograms. all systematics that can be included are. ML is not yet included.


ekauffma:
my output of pip list: 
[env.txt](https://github.com/iris-hep/analysis-grand-challenge/files/11882558/env.txt)

PR #165, opened 2023-06-27
feat: save histograms without re-binning

alexander-held:
This addresses the second item in #162 via an intermediate solution that can be replaced once https://github.com/scikit-hep/cabinetry/issues/412 is addressed. It makes use of https://cabinetry.readthedocs.io/en/latest/advanced.html#overrides-for-template-building to subsequently re-bin via `cabinetry`, which ends up less complicated that I had initially expected.

The histograms getting saved to disk now are no longer re-binned from the coffea output, so they're 25 bins from 50 to 550 GeV (20 GeV bin width). During template collection in `cabinetry` the lowed bin edge gets set to 110 GeV and the bins are merged 2->1, which previously was already done to the histograms saved to the root file.


ekauffma:
Looks fine to me and runs on my end!

ekauffma (approved):

Issue #168, opened 2023-06-28
Histogram validation and bin migrations


alexander-held:
In the CMS ttbar setup there is a histogram validation script (and a reference) provided. There are cases where the observable calculated for an event seems to be extremely close to a bin boundary, and in such cases the event may migrate between bins. This is presumably due to floating point math, and can happen even with the same implementation when using different machines. It can look like this:

```
4j2b_single_top_t_chan_pt_scale_up
        Contents do not match:
        got      [18.34436054206814, 191.94135120855094, 616.3569931138268, 1158.3923520829348, 1829.0405553414591, 2620.8829280746218, 3549.76845738687, 3981.8725441002866, 3059.056839636495, 2392.6575137718214, 2143.929579980845, 1915.501838183115, 1729.900082816315, 1598.0499986077054, 1437.671737611248, 1326.0544028975658, 1230.6907102781418, 1101.8080959954548, 1029.7795077935136, 962.6742219686713, 882.1478789788599, 816.2565581237128, 752.7931672079566, 705.7857458814076, 652.1689597198491, 593.0218886963431, 9634.026000505984]
        expected [18.34436054206814, 191.94135120855094, 616.3569931138268, 1158.3923520829348, 1829.0405553414591, 2620.8829280746218, 3549.76845738687, 3981.8725441002866, 3059.056839636495, 2392.6575137718214, 2143.8621374825284, 1915.5692806814316, 1729.900082816315, 1598.0499986077054, 1437.671737611248, 1326.0544028975658, 1230.6907102781418, 1101.8080959954548, 1029.7795077935136, 962.6742219686713, 882.1478789788599, 816.2565581237128, 752.7931672079566, 705.7857458814076, 652.1689597198491, 593.0218886963431, 9634.026000505984]
```

where an event migrates between two bins:
```
observed: 2143.929579980845, 1915.501838183115
reference: 2143.8621374825284, 1915.5692806814316
```

In such cases the tolerance can be increased to compensate, but we might want to have a better method to spot this. Importantly, partial sums of the counts should still match in case of migrations.

cc @eguiraud @ekauffma 

Some details also in https://github.com/iris-hep/analysis-grand-challenge/pull/163#issuecomment-1608403618.

ekauffma:
Should this be closed by #171 or is there further work to be done here?

alexander-held:
Indeed this is addressed by #171, thanks for following up here.

PR #171, opened 2023-06-29
feat: updated validation script to check partial bin sums

ekauffma:
This is an attempt to address #168 . If adjacent bins are similar enough, the partial sum of the affected bins is compared. This will ensure that one event migrating across bins will not flag as an error. Maybe we can test this with RDF implementation (@andriiknu)

@alexander-held @eguiraud

### DIFF
--- a/analyses/cms-open-data-ttbar/validate_histograms.py
+++ b/analyses/cms-open-data-ttbar/validate_histograms.py
@@ -73,7 +73,7 @@ def validate(histos: dict, reference: dict, verbose=False) -> dict[str, list[str
                 h_group = np.array(h['contents'])[group]
                 ref_group = np.array(ref_h['contents'])[group]
                 # if difference is great, count as error
-                if not np.allclose(h_group, ref_group, atol=2.0):
+                if not np.allclose(h_group, ref_group, atol=2.0): # 2.0 is chosen as it seems to cover the difference expected of one events migrating between bins in histograms from 1 file per sample. this definitely can be tuned in the future
                     is_error = True
                     if verbose:
                         print(f"In {name}: Not close enough for bin migration")
--- a/analyses/cms-open-data-ttbar/validate_histograms.py
+++ b/analyses/cms-open-data-ttbar/validate_histograms.py
@@ -43,12 +43,11 @@ def validate(histos: dict, reference: dict, verbose=False) -> dict[str, list[str
         if not np.allclose(h['edges'], ref_h['edges']):
             errors[name].append(f"Edges do not match:\n\tgot      {h['edges']}\n\texpected {ref_h['edges']}")
         contents_depend_on_rng = "pt_res_up" in name # skip checking the contents of these histograms as they are not stable
+        is_close = np.isclose(h['contents'], ref_h['contents']) 
+        
+        #### check if bin migration ####
+        if not contents_depend_on_rng and not all(is_close):
 
-        if not contents_depend_on_rng and not np.allclose(h['contents'], ref_h['contents']):
-
-            #### check if bin migration ####
-            # returns boolean array that returns false elementwise if not close
-            is_close = np.isclose(h['contents'], ref_h['contents']) 
             # gets indices where above array is false
             where_not_close = np.where(np.invert(is_close))[0]
             # gets difference of adjacent entries
--- a/analyses/cms-open-data-ttbar/validate_histograms.py
+++ b/analyses/cms-open-data-ttbar/validate_histograms.py
@@ -65,17 +65,14 @@ def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
                 ref_group = np.array(ref_h['contents'])[group]
                 # if difference is great, count as error
                 if not np.allclose(h_group, ref_group, atol=5e-1):
-                    print("Adjacent bins have difference which is too great:")
-                    print("histogram: ", h_group, ", reference: ", ref_group)
-                    print()
                     is_error = True
                 # check partial sum
                 if not np.allclose(sum(h_group), sum(ref_group)):
-                    print("Partial sum is not equal:")
+                    is_error = True
+                else:
+                    print("Bin migration likely:")
                     print("histogram: ", h_group, ", reference: ", ref_group)
                     print()
-                    is_error = True
-
             if is_error:
                 errors[name].append(f"Contents do not match:\n\tgot      {h['contents']}\n\texpected {ref_h['contents']}")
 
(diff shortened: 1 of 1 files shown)


analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -42,8 +42,39 @@ def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
...
         contents_depend_on_rng = "pt_res_up" in name # skip checking the contents of these histograms as they are not stable
+
         if not contents_depend_on_rng and not np.allclose(h['contents'], ref_h['contents']):
-            errors[name].append(f"Contents do not match:\n\tgot      {h['contents']}\n\texpected {ref_h['contents']}")
+
+            # check if bin migration
+            is_close = np.isclose(h['contents'], ref_h['contents'])
+            where_not_close = np.where(np.invert(is_close))
+            diff_values = np.diff(where_not_close)
+            split_indices = np.argwhere(np.abs(diff_values[0])>1)
```
eguiraud, line 52: I think the `[0]` here logically belongs at line 50, where `np.where` returns an extra dimension that we already don't care about at line 51. removes the need to repeat the `[0]` at the lines below too.

analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -42,8 +42,39 @@ def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
...
+            split_indices_mod = []
+            for i in range(len(split_indices)):
+                if len(split_indices[i])>0:
+                    split_indices_mod.append(split_indices[i][0]+1)
+            if len(split_indices_mod)>0:
+                split_values = np.split(where_not_close[0], split_indices_mod)
+            else:
+                split_values = [where_not_close[0]]
```
eguiraud, line 60: at this point I was expecting that if `split_indices` is not empty (so non-adjacent bins differ from the reference) we'd be done: they cannot be due to a bin migration (I wouldn't expect values to migrate between non-adjacent bins). I am not sure what `split_indices_mod` is/does :sweat_smile: 

also we only have 1D histograms so `len(split_indices)` will always be 1 I think?

analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -42,8 +42,39 @@ def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
...
+                # if difference is great, count as error
+                if not np.allclose(h_group, ref_group, atol=5e-1):
+                    is_error = True
+                # check partial sum
+                if not np.allclose(sum(h_group), sum(ref_group)):
+                    is_error = True
+                else:
+                    print("Bin migration likely:")
+                    print("histogram: ", h_group, ", reference: ", ref_group)
+                    print()
```
eguiraud, line 75: maybe we can hide this behind a verbose flag: as bin migrations are not an issue, we should not "complain" to users about them, they might be confused/think there is a problem.

analyses/cms-open-data-ttbar/validate_histograms.py
eguiraud, line 60: It's likely that I'm just not good enough at numpy/too dumb, but the lines above are quite dense and I had to follow along in an interpreter, could not do the operations in my head. consider adding comments that show the reader what's going on.

eguiraud (commented):
thanks @ekauffma , I would have approved the PR but I don't have enough permission to do so, I can only comment :smile: 

I could not completely follow the logic in my head, so sorry if some of the comments don't make sense. Even with the comments, that you can feel free to disregard, this overall LGTM.

ekauffma:
@eguiraud I made a lot of edits inspired by your suggestions. I would appreciate if you would read again and let me know if you find it easier to follow now! Thank you

analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -42,8 +43,57 @@ def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
...
+        if not contents_depend_on_rng and not all(is_close):
+
+            # gets indices where above array is false
+            where_not_close = np.where(np.invert(is_close))[0]
+            # gets difference of adjacent entries
+            diff_values = np.diff(where_not_close)
+            # get the indices where the above difference is greater than 1 (to form groupings of bins based on location)
+            split_indices = np.argwhere(np.abs(diff_values)>1)
+            # np.argwhere adds extra dimension we don't need
+            split_indices = split_indices.reshape((split_indices.shape[0],))
```
eguiraud, line 58: is this the same as `split_indices = split_indices[0]`?

analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -42,8 +43,57 @@ def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
...
+                # split indices into groupings
+                split_values = np.split(where_not_close, split_indices)
+            
+            is_error=False
+            # iterate through groupings and test for bin migration or error
+            for group in split_values:
+                h_group = np.array(h['contents'])[group]
+                ref_group = np.array(ref_h['contents'])[group]
+                # if difference is great, count as error
+                if not np.allclose(h_group, ref_group, atol=2.0):
```
eguiraud, line 76: `2.0` looks magical, maybe add a not about where it comes from (e.g. point to the GitHub issues where there are measurements on the impact of bin migrations)

eguiraud (commented):
Looks great!! Took me less than 10 minutes to go through. What I didn't understand yesterday was "(to form groupings of bins based on location)" :grimacing: :sweat_smile:

analyses/cms-open-data-ttbar/validate_histograms.py
alexander-held: We briefly talked about this today as well, it is indeed a pretty magical number and if we have a reference that would be useful. It is most strict with very few files (assuming migrations are rare to begin with), as with more files each event will have a smaller weight and it would require the same proportion of additional events migrating to stick with the same absolute migration. I assume that people would mostly compare setups with few files, for which a value of 2 doesn't feel unreasonable.

eguiraud: That's totally ok, I was just suggesting to add a comment to explain this (or point to an explanation) to someone reading the code, because they might need to know where the 2. came from to decide whether it still makes sense in 5 years or for a slightly different use case

alexander-held: For sure, I agree with the comment.

alexander-held (approved):
Looks good to me, thanks for adding this!

PR #175, opened 2023-07-19
feat: updated validation script to check partial bin sums

eguiraud:
* updated validation script to check partial bin sums
* added verbose flag

This is a backport of #171 .

### DIFF
--- a/analyses/cms-open-data-ttbar/validate_histograms.py
+++ b/analyses/cms-open-data-ttbar/validate_histograms.py
@@ -93,7 +93,6 @@ def validate(histos: dict, reference: dict, verbose=False) -> dict[str, list[str
                         print()
             if is_error:
                 errors[name].append(f"Contents do not match:\n\tgot      {h['contents']}\n\texpected {ref_h['contents']}")
-            print()
 
     return errors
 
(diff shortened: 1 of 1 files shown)


analyses/cms-open-data-ttbar/validate_histograms.py
```diff
@@ -42,8 +43,57 @@ def validate(histos: dict, reference: dict) -> dict[str, list[str]]:
...
+                        print("histogram: ", h_group, ", reference: ", ref_group)
+                        print()
+                else:
+                    if verbose:
+                        print(f"In {name}: Bin migration likely")
+                        print("histogram: ", h_group, ", reference: ", ref_group)
+                        print()
+            if is_error:
+                errors[name].append(f"Contents do not match:\n\tgot      {h['contents']}\n\texpected {ref_h['contents']}")
+                print()
```
alexander-held, line 96: I see we also have it in `main`, but isn't this `print()` completely superfluous? It only fills up the output with blank lines in case of disagreements.

eguiraud: I think you are right, let me remove it. I have to fix #176 as well

eguiraud: removed

alexander-held: Thanks!

PR #177, opened 2023-07-19
feat: do not require pseudodata hists in validate_histograms.py

eguiraud:
These histograms are needed for the statistical inference stage, but they are built from other histograms and it is up to the implementation whether to write out dedicated pseudodata histograms or build them on the fly.

If approved I'll backport to the v1 branch.


alexander-held (approved):
Looks good, thank you!

PR #180, opened 2023-07-20
docs: add v1.2 description and fix cut descriptions

alexander-held:
Adding a description of the v1.2 tag, fixing the jet eta cut and explaining how to obtain the tight lepton veto via the variable provided in the input files. Also clarify how to pick jets for the ML part if not considering all permutations.


ekauffma (approved):
Looks good to me!

alexander-held:
Thanks for checking @ekauffma!

PR #183, opened 2023-07-25
docs: Add columnflow implementation example of AGC

matthewfeickert:
* c.f. https://github.com/columnflow/agc_cms_ttbar

cc @riga


alexander-held (approved):
Thanks for this addition!

alexander-held:
I was too fast to approve and forgot that we also list this in the docs section, fixed by #184.

PR #188, opened 2023-08-19
docs: clarify ServiceX setup instructions

TatianaOvsiannikova:


analyses/atlas-open-data-hzz/HZZ_analysis_pipeline.py
```diff
@@ -88,8 +88,8 @@
 # %% [markdown]
 # ### Required setup for this notebook
 #
-# If you are running on coffea-casa, you should be good to go: ServiceX credentials are automatically available to you.
-# If not, you will need to set those up.
+# If you are running on coffea-casa with CMS account ServiceX credentials are automatically available to you.
+# Otherwise you will need to set those up.
```
alexander-held, line 92: Thanks for the text update! A small clarification: the coffea-casa instance we should refer to here is the Open Data instance. That one does not require a CMS account. There is also a CMS instance for CMS members, but we do not need to mention it here.

While we're updating this text anyway, it might also be useful to add a link to the ServiceX docs for some more information about setting this up manually. I believe the best link for that is https://servicex.readthedocs.io/en/latest/user/getting-started/

analyses/atlas-open-data-hzz/HZZ_analysis_pipeline.py
```diff
@@ -88,11 +88,11 @@
 # %% [markdown]
 # ### Required setup for this notebook
 #
-# If you are running on coffea-casa with CMS account ServiceX credentials are automatically available to you.
+# If you are running on coffea-casa Open Data instance with CMS account ServiceX credentials are automatically available to you.
```
alexander-held, line 91:
```suggestion
# If you are running on the coffea-casa Open Data instance, ServiceX credentials are automatically available to you.
```

alexander-held (approved):

PR #191, opened 2023-08-30
feat: move xcache prefix setting to config

alexander-held:
This extends #189 to store the xcache prefix setting in the benchmarking config.


oshadura (approved):
LGTM! :)

PR #192, opened 2023-09-01
Updating client code for obtaining Dask client on EAF

Nanoemc:
The code was reverted back to lpcgateway, which is outdated for the EAF. This commit updates clients.py to be compatible with the EAF@FNAL.


alexander-held (approved):
Thanks for keeping this up to date!

PR #200, opened 2023-09-07
feat: extend metrics to support more file sources

alexander-held:
Better tracking of file origin in metrics and a lightweight way to distinguish between v1 and main.


oshadura (commented):
Can you add please also in list "red-xcache1.unl.edu:1096", which is our OD xcache?
Otherwise LGTM!

oshadura (approved):

PR #202, opened 2023-09-07
feat: track inference settings

alexander-held:
Tracking the inference on/off setting and whether Triton is used.


oshadura (approved):
LGTM!

PR #204, opened 2023-09-07
fix: avoid column overtouching in ML input feature calculation

alexander-held:
The previous calculation of ML input features accidentally caused the full lepton information to get materialized. This results in a significant I/O increase and slowdown. This update avoids the overtouching by performing the relevant piece of the calculation with a minimal re-built set of four-vectors with all other information stripped out.

This is expected to have significant performance impact for I/O bound setups.

see https://github.com/CoffeaTeam/coffea/issues/892 for details


oshadura (approved):
LGTM! If you merge it ASAP, I can try to do the final measurements with it..

PR #208, opened 2023-09-13
feat: do not delete trained models by default

alexander-held:
Keeps around trained models by default. Also add instructions for downloading `mc` for model uploading.


oshadura (approved):
LGTM!

PR #211, opened 2023-09-22
Modify the CMS tt-bar analysis to work with the new ServiceX client

BenGalewsky:
In doing this, discovered a few issues with the current version of the library so I published Alpha 6 of the library.

This change resulted in a much cleaner ServiceX interaction. I didn't test out the Coffea portion, but  this should work.


alexander-held (commented):
Thanks a lot for porting this! I'm looking forward to the demo. It would be good to hear the plans for the full release of the ServiceX frontend v3.0, as I would like to wait with merging this until we have a full release available.

PR #213, opened 2023-11-16
feat: improve handling of optional dependencies

alexander-held:
Refactor imports to allow running without ServiceX client (and related libraries) installed. Also avoid importing everything into `utils` to ensure `cabinetry` is not needed on worker nodes (which import `utils` at the top level).


oshadura (approved):
LGTM! Works for locally! I would like to merge to check on the new setup we have at UNL.

PR #214, opened 2023-11-30
refactor: save available histograms and avoid hardcoding names

alexander-held:
To avoid hardcoding too much in the histogram saving function, instead look at what is contained in the histogram object and save everything that is found in the right format.


oshadura (approved):
I like that code become so much easier to read! LGTM!

PR #216, opened 2024-04-11
Update config.py

rasmalais:
Change  "CORES_PER_WORKER" to 1 from 2.


analyses/cms-open-data-ttbar/utils/config.py
```diff
@@ -28,7 +28,7 @@
         # currently has no effect
         "SYSTEMATICS": "all",
         # does not do anything, only used for metric gathering (set to 2 for distributed coffea-casa)
```
alexander-held, line 30:
```suggestion
        # does not do anything, only used for metric gathering (set to 1 for distributed coffea-casa)
```

alexander-held (approved):
Thanks a lot for spotting and fixing this @rasmalais! We changed the setup last year but we did not remember to correct the default value here, so this is very good to have fixed now.

PR #217, opened 2024-05-17
REANA set up

AndriiPovsten:
The set up which is required to run the AGC on the REANA platform.
-  Repr-step(REANA) folder
-  Modified `ttbar_analysis_reana.ipynb`
-  required `file/final_merging.ipynb`
- `Snakefile`
- `reana.yaml file`


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -1,46 +1,33 @@
...
-#       format_version: '1.3'
-#       jupytext_version: 1.15.2
-#   kernelspec:
-#     display_name: Python 3 (ipykernel)
-#     language: python
-#     name: python3
-# ---
-
-# %% [markdown]
+#!/usr/bin/env python
```
alexander-held, line 1: We should do the same jupytext re-sync like in the HEPData case.

AGC_clean/analysis-grand-challenge
alexander-held, line 1: This submodule should be removed.

analyses/cms-open-data-ttbar/utils/config.py
```diff
@@ -18,7 +18,7 @@
         # read files from public EOS (thanks to the CMS DPOA team!)
         # note that they are likely only available temporarily
         # and not part of an official CMS Open Data release
-        "INPUT_FROM_EOS": False,
+        "INPUT_FROM_EOS": True,
```
alexander-held, line 21: We shouldn't change the default here.

analyses/cms-open-data-ttbar/utils/file_input.py
```diff
@@ -158,4 +158,4 @@ def get_data_rootfiles_uri(self, query, as_signed_url=True, title="Untitled"):
             # update files for each process
             files_per_process.update({process: all_files[parent_key[self.filelist[:,1]==process]]})
 
-        return files_per_process
+        return files_per_process
```
alexander-held, line 161: Same here as for HEPData: let's not change the newline to minimize files being touched.

analyses/cms-open-data-ttbar/Repr-step(REANA)/file_merging.ipynb
alexander-held, line 1: The merging steps should probably be a script instead of a notebook?

AndriiPovsten:
@alexander-held  the only thing which would require for successful run AGC on REANA is to update the harbor image with e `papermill` package, I'm not sure how it can be done. I could ask Matthew about it.

analyses/cms-open-data-ttbar/Repr-step(REANA)/file_merging.ipynb
AndriiPovsten: In theory, the workflow can be executed as a script, but the `Snakefile` would be much longer, I believe executing notebooks is sufficient.

PR #219, opened 2024-07-04
feat: add a client for Purdue Analysis Facility

kondratyevd:
@oshadura @alexander-held following up on our chat last month:

This PR adds an option to run CMS OpenData workflow at Purdue Analysis Facility via Dask Gateway cluster.
The setup is similar to the `cmsaf-dev` option, the only differences are:

- The address of the Dask Gateway API server;
- The default number of workers is 10 to avoid using too much resources when running AGC as a "smoke test".

Additional info about the Dask Gateway setup at Purdue AF can be found [here](https://analysis-facility.physics.purdue.edu/en/latest/doc-dask-gateway.html). At Purdue AF, we have two Dask Gateway instances - with SLURM and Kubernetes backends. The `purdue-af` AGC client will use the Kubernetes backend.

### DIFF
--- a/analyses/cms-open-data-ttbar/utils/clients.py
+++ b/analyses/cms-open-data-ttbar/utils/clients.py
@@ -27,7 +27,18 @@ def get_client(af="coffea_casa"):
         cluster.scale(50)
         
         client = cluster.get_client()
-        
+
+    elif af == "purdue-af":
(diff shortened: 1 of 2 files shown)


oshadura (approved):
LGTM!

alexander-held (approved):
Thanks for adding this!

PR #225, opened 2024-09-12
CMS example with ServiceX 3 client

ponyisi:
This works with uproot-raw right now with `servicex==3.0.0b1`. The func_adl version does not yet work (@gordonwatts is working on this), and in the end we would want a version with the final 3.0.0 release.


analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -192,9 +194,12 @@ def process(self, events):
...
             xsec_weight = 1
 
-        # setup triton gRPC client
-        if self.use_inference and self.use_triton:
-            triton_client = utils.clients.get_triton_client(utils.config["ml"]["TRITON_URL"])
+        # setup triton gRPC client or xgboost
+        if self.use_inference:
+            if self.use_triton:
+                triton_client = utils.clients.get_triton_client(utils.config["ml"]["TRITON_URL"])
+            else:
```
alexander-held, line 201: Is this some leftover from debugging xgboost?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -473,32 +509,30 @@ def get_query(source):
...
-    # create the query
-    query = get_query(dummy_ds).value()
+    if USE_SERVICEX_UPROOT_RAW:
+        hl_query = get_uproot_raw_query()
+    else:
+        hl_query = get_query(query.FuncADL_Uproot())
 
     # now we query the files using a wrapper around ServiceXDataset to transform all processes at once
     t0 = time.time()
-    ds = utils.file_input.ServiceXDatasetGroup(fileset, backend_name="uproot", ignore_cache=utils.config["global"]["SERVICEX_IGNORE_CACHE"])
```
alexander-held, line 494: Since I think we can fully get rid of `ServiceXDatasetGroup` now, we can also drop it from the `utils` module.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
```diff
@@ -465,6 +470,37 @@ def get_query(source):
                                        "Jet_jetId": h.Jet_jetId,
                                       })
 
+def get_uproot_raw_query():
+    cut = '((count_nonzero((Electron_pt > 30) & (abs(Electron_eta) < 2.1) & (Electron_cutBased == 4) & (Electron_sip3d < 4), axis=1)' \
```
alexander-held, line 474: Is there a good way to write this in a way that is not a string for syntax highlighting purposes? I was hoping one could write this as a function taking a tree and returning an awkward array, allowing to also easily debug this locally by feeding a tree from uproot.

ponyisi: Unfortunately this is not at all trivial to do (if you really want to do that what you want is the Python codegen instead). In some later release we can add a capability to reasonably easily test it locally with `uproot` calls. (It's not trivial at the moment because we have to manually bootstrap awkward functions into the uproot execution space.)

At least JupyterLab does highlighting of the parentheses in the strings, so it's actually not so terrible to write the expressions correctly.

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
ponyisi: ? This is mandatory if you use the `FuturesExecutor` : the processes will lock if you attempt to share an xgboost model from the originating processes, and best practice suggests they be loaded inside the daughters anyway. For Dask this should have no meaningful effect since there are no copy-on-write benefits to be had anyway.

ponyisi:
I've pushed the fixes for the func_adl version of the CMS notebook, with fixes courtesy of @gordonwatts (this needs func_adl >= 3.3.3).

I would like to update the ATLAS example as well, but it seems that at the moment there are problems with the UNL storage which prevent me from accessing those files.

ponyisi:
I have updated the ATLAS examples.

analyses/atlas-open-data-hzz/HZZ_analysis_pipeline.py
```diff
@@ -179,23 +181,35 @@ def get_lepton_query(source: ObjectStream) -> ObjectStream:
...
 # Using the queries created with `func_adl`, we are using `ServiceX` to read the ATLAS Open Data files to build cached files with only the specific event information as dictated by the query.
 
 # %%
-# dummy dataset on which to generate the query
-dummy_ds = ServiceXSourceUpROOT("cernopendata://dummy", "mini", backend_name="uproot")
-
-# tell low-level infrastructure not to contact ServiceX yet, only to
-# return the qastle string it would have sent
-dummy_ds.return_qastle = True
+import servicex
```
alexander-held, line 205: For consistency with all the other package imports, could we move this to the top of the file please?

analyses/cms-open-data-ttbar/ttbar_analysis_pipeline.py
alexander-held: I see, we probably haven't run that executor in quite a while then. Thanks!

alexander-held:
The issue with the CI complaining about unrelated files should be fixable by rebasing to incorporate #226.

analyses/cms-open-data-ttbar/utils/file_input.py
```diff
@@ -119,43 +110,3 @@ def download_file(url, out_file):
     out_path.parent.mkdir(parents=True, exist_ok=True)
     with tqdm.tqdm(unit='B', unit_scale=True, unit_divisor=1024, miniters=1, desc=out_path.name) as t:
         urllib.request.urlretrieve(url, out_path.absolute(), reporthook=tqdm_urlretrieve_hook(t))
-
-
-class ServiceXDatasetGroup():
```
alexander-held, line 124: `numpy` is no longer needed here with `ServiceXDatasetGroup` removed (this is about line 2 but GitHub won't let me comment there as it's not part of the diff)

alexander-held:
In the CMS ttbar notebook after the ServiceX delivery cell I see

> `Method Zip not found on object <class 'types.dict_dataclass'>`

What is this about?

alexander-held:
Other than the small things about imports above, could we please keep ServiceX off by default with `USE_SERVICEX = FALSE` to make the notebook work by default with the least amount of dependencies?

One more very picky request: could we please move `USE_SERVICEX_DOWNLOAD` into `utils/config.py` (probably into the `global` section) in an attempt to limit the amount of things to be configured in the notebook directly?

analyses/cms-open-data-ttbar/requirements.txt
```diff
@@ -1,169 +1,200 @@
...
+rpds-py==0.20.0
 rsa==4.9
+ruamel.yaml==0.18.6
+ruamel.yaml.clib==0.2.8
+scikit-learn==1.5.2
 scipy==1.10.1
-Send2Trash==1.8.0
-servicex==2.6.2
+Send2Trash==1.8.3
+servicex==3.0.0b1
```
alexander-held, line 161: This can now be updated to the released version.

ponyisi:
> In the CMS ttbar notebook after the ServiceX delivery cell I see
> 
> > `Method Zip not found on object <class 'types.dict_dataclass'>`
> 
> What is this about?

This is a particular issue related to how `func_adl` interprets the query, I know @gordonwatts is looking into it.

alexander-held (approved):
This looks all good to me, thank you very much! I believe @oshadura wanted to ensure we can run this on the Open Data coffea-casa before merging but I'll let her comment.

PR #233, opened 2026-05-15
Update ATLAS analysis notebook to updated software 

ArturU043:
Similar issues as in #232 
1) **Fixed the order of arguments in the coffea Runner with explicit parsing**, which was causing crashes

2) Additionally, workers were getting 403 when trying to read ServiceX output from the S3 object store - to investigate. 
Changing the delivery type to `LocalCache` will ensure the AGC pipeline always works. 

3) Minor adjustments to avoid warnings in strings using LaTeX by adding r-strings.


alexander-held (approved):
Looks good to me, thanks!

